PR #7, opened 2019-03-26
Add subcommands for wp-cli, logs, stopping, and status

joehoyle:
Change the local-server to use subcommands: start, stop, status, cli, logs

Help section now:

```
Help:
  Run the local development server.

  To start the local development server:
  	start
  Stop the local development server:
  	stop
  View status of the local development server:
  	status
  Run WP CLI command:
  	cli -- <command>              eg: cli -- post list --debug
  View the logs
  	logs <service>                <service> can be php, nginx, db, s3, elasticsearch, xray
```

There's a bit of a niggle on how the cli arguments have to be passed. you can do `cli posts list`, but you have to do `cli -- posts list --url=some-site`, otherwise the `--url` option will be parsed by composer and fail.

### DIFF
--- a/inc/Composer/Command.php
+++ b/inc/Composer/Command.php
@@ -3,32 +3,131 @@
 namespace HM\Platform\LocalServer\Composer;
 
 use Composer\Command\BaseCommand;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Process\Process;
 
 class Command extends BaseCommand {
 	protected function configure() {
-		$this->setName( 'local-server' );
+		$this->setName( 'local-server' )
+			->setDescription( 'Local server.' )
+			->setDefinition( [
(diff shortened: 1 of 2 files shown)


roborourke:
That niggle is the same deal as Pantheon's terminus so not that much of an issue. Cheers for this

inc/Composer/Command.php
```diff
@@ -3,32 +3,131 @@
...
+
+		$site_url = 'https://' . basename( getcwd() ) . '.altis.dev/';
+		$output->writeln( 'Startup completed.' );
+		$output->writeln( 'To access your site visit: ' . $site_url );
+	}
+
+	protected function stop( InputInterface $input, OutputInterface $output ) {
+		$output->writeln( 'Stopping...' );
+
+		$proxy = new Process( 'docker-compose down', 'vendor/humanmade/local-server/docker' );
```
roborourke: Does this exit with an error if it's already stopped?

joehoyle: I think it's docker-compose is happy with it

joehoyle:
@roborourke if it were being passed untrsuted data maybe, but given we explicitly want to forward all options, it's better not to

Issue #8, opened 2019-03-28
Traefik redirects to wrong backend


nathanielks:
When there are two sites, foo.example.dev and bar.example.dev, and submitting 2 requests to foo.example.dev, one of the requests will be served by bar.example.dev's backend but the url will be foo.example.dev. I've noticed this locally but haven't logged it anywhere, but it's something to figure out!

nathanielks:
Another way of saying this is running multiple local servers at the same time won't work as expected as every 1/N requests will make it to the correct environment, where N is the number of environments currently running.

joehoyle:
Hmm interested, I thought Traefik was set up with labels to direct the correct hostname to the correct container, or am I missing something?

nathanielks:
Nope, you're assumption is correct! That's how I believe it's to be working, but it doesn't appear so!

nathanielks:
I think I've figured it out. It may not be Traefik, after all. I've switched to using https://github.com/jwilder/nginx-proxy as that seems like a simpler implementation and I'm getting the same error. It then occurred to me that we're specifying backend as simply `php`. Because multiple environments are all on the same docker network, `php` refers to N backends, where is the number of environments running. Next up is to figure out how to get each `nginx` container to reference _its_ `php` container.

nathanielks:
A simple test on the `nginx` container:
```
# curl -vv php
* Rebuilt URL to: php/
*   Trying 172.21.0.7...
* TCP_NODELAY set
* connect to 172.21.0.7 port 80 failed: Connection refused
*   Trying 172.21.0.13...
* TCP_NODELAY set
* connect to 172.21.0.13 port 80 failed: Connection refused
* Failed to connect to php port 80: Connection refused
* Closing connection 0
curl: (7) Failed to connect to php port 80: Connection refused
```

`php` has multiple definitions, so it attempts to resolve each of them in the order they're defined.

nathanielks:
Aha! Brilliant! The trick is to have _only_ the `nginx` container on the proxy network and then use the default network for everything, including the `nginx` container. This means the `nginx` container for the environment is both on the proxy network as well as the default network for that environment, meaning `php` will resolve to the `php` container on the default network 👍 💯

roborourke:
Can the name `proxy` be dynamic eg. `"${COMPOSE_PROJECT_NAME}-proxy"`?

nathanielks:
@roborourke nope, because there's a single instance of the proxy running:

[...]

Issue #2, opened 2019-03-21
Expose ES port


roborourke:
In a similar vein to #1 you often need to connect to and play with ES via the head extension for chrome or similar to debug and develop queries.

roborourke:
Elasticsearch is accessible at `elasticsearch-<project-name>.altis.dev`

Issue #1, opened 2019-03-21
Add database connection helper command


roborourke:
It's easy to connect to a chassis db for example so it'd be useful to be able to do the same here. The command should be named `composer server db`.

Acceptance criteria:

- [x] `composer server db` should log the user into mysql on the db server
- [x] `composer server db spf` should generate and open an SPF file for sequel pro (also add an alias of `sequel`)
- [x] `composer server db info` should print out db connection details
- [x] Update command help output
- [x] Document db command in a new file `docs/database.md`

`composer server db info` output template:

```
Root password: wordpress

Database: wordpress
Username: wordpress
Password: wordpress

Host: 0.0.0.0
Port: 31232

Version: MySQL 5.7
```

Bonus

- [ ] `composer server db workbench` Windows support for MySQL Workbench

roborourke:
This one is a bit meatier but will be helpful. A few more notes to assist in development:

- See the `composer server logs` command for a reference on how to check for subcommands and arguments
- Example `.spf` template: https://github.com/Chassis/SequelPro/blob/master/template.spf
- Get the mapped db ports using
  ```
  docker ps --format "{{.Ports}}" --filter name=altis-dev_db_1
  # -> 33060/tcp, 0.0.0.0:32123->3306/tcp
  ```
  `0.0.0.0` is the host and `32123` in this case is the port on the host you can access it at
- username and pass is `wordpress`, `wordpress`

roborourke:
Closed by #122

PR #9, opened 2019-04-01
Remove unneeded proxy network from services not needing to be exposed

nathanielks:
Fixes https://github.com/humanmade/local-server/issues/8.


docker/docker-compose.yml
```diff
@@ -28,7 +25,7 @@ x-php: &php
     PAGER: more
     HM_ENV_ARCHITECTURE: 'local-server'
     HM_DEPLOYMENT_REVISION: 'dev'
-    ELASTICSEARCH_HOST: elasticsearch
+    ELASTICSEARCH_HOST: elasticsearch.${COMPOSE_PROJECT_NAME:-default}.altis.dev
     ELASTICSEARCH_PORT: 9200
     AWS_XRAY_DAEMON_HOST: xray
     S3_UPLOADS_ENDPOINT: http://s3.localhost:8000/
```
nathanielks: @joehoyle 	I can't find any reference to `S3_UPLOADS_ENDPOINT` in S3 Uploads. How/where is this used? I'm working on updating references to services in a way that won't cause issues like #8 again.

nathanielks:
The Elasticsearch image is failing to build for me. Is this happening to you too, @joehoyle?

docker/docker-compose.yml
joehoyle: It's not automatically used, see https://github.com/humanmade/local-server/blob/master/load.php#L26 for an example of it being used though

docker/docker-compose.yml
```diff
@@ -28,7 +25,7 @@ x-php: &php
     PAGER: more
     HM_ENV_ARCHITECTURE: 'local-server'
     HM_DEPLOYMENT_REVISION: 'dev'
-    ELASTICSEARCH_HOST: elasticsearch
+    ELASTICSEARCH_HOST: elasticsearch.${COMPOSE_PROJECT_NAME:-default}.altis.dev
```
joehoyle: Shouldn't hte ES instance still be addressable on the elasticseearch (only) hostname?

joehoyle:
@nathanielks what error are you getting?

docker/docker-compose.yml
nathanielks: If we do that, we run into the same problem as before where more than 1 elasticsearch container IP will be assigned to that hostname because they all exist on the same network (proxy). If the elasticsearch service doesn't need to be addressable outside of it's environment, then we can remove the Traefik labels and address it via the `elasticsearch`. Is that clear as mud? 😛

joehoyle: Ohh gotcha, forgot we were putting it on the proxy network. I think we _do_ want it externally addressable, so that's 👍 then.

nathanielks:
@joehoyle I hadn't dug into it at all in case you were already aware and working on it!

---

```
$ VOLUME=~/Work/repos/hm-base COMPOSE_PROJECT_NAME=hm-base dc up                                                                                                           9h22m  fix-backend-routing [319dcb7]
Creating network "hm-base_default" with the default driver
Building elasticsearch
Step 1/2 : FROM blacktop/elasticsearch:6.3
 ---> 69b8ba374ac9
Step 2/2 : RUN bin/elasticsearch-plugin install --batch --silent ingest-attachment
 ---> Running in 5b2ac241ab1c
Exception in thread "main" java.lang.ExceptionInInitializerError
	at sun.security.ssl.SSLSessionImpl.<init>(SSLSessionImpl.java:188)
	at sun.security.ssl.SSLSessionImpl.<init>(SSLSessionImpl.java:152)
	at sun.security.ssl.SSLSessionImpl.<clinit>(SSLSessionImpl.java:79)
	at sun.security.ssl.SSLSocketImpl.init(SSLSocketImpl.java:598)
	at sun.security.ssl.SSLSocketImpl.<init>(SSLSocketImpl.java:536)
	at sun.security.ssl.SSLSocketFactoryImpl.createSocket(SSLSocketFactoryImpl.java:72)
	at sun.net.www.protocol.https.HttpsClient.createSocket(HttpsClient.java:405)
	at sun.net.NetworkClient.doConnect(NetworkClient.java:162)
	at sun.net.www.http.HttpClient.openServer(HttpClient.java:463)
	at sun.net.www.http.HttpClient.openServer(HttpClient.java:558)
	at sun.net.www.protocol.https.HttpsClient.<init>(HttpsClient.java:264)
	at sun.net.www.protocol.https.HttpsClient.New(HttpsClient.java:367)
	at sun.net.www.protocol.https.AbstractDelegateHttpsURLConnection.getNewHttpClient(AbstractDelegateHttpsURLConnection.java:191)
	at sun.net.www.protocol.http.HttpURLConnection.plainConnect0(HttpURLConnection.java:1156)
	at sun.net.www.protocol.http.HttpURLConnection.plainConnect(HttpURLConnection.java:1050)
	at sun.net.www.protocol.https.AbstractDelegateHttpsURLConnection.connect(AbstractDelegateHttpsURLConnection.java:177)
	at sun.net.www.protocol.https.HttpsURLConnectionImpl.connect(HttpsURLConnectionImpl.java:162)
	at org.elasticsearch.plugins.InstallPluginCommand.urlExists(InstallPluginCommand.java:325)
	at org.elasticsearch.plugins.InstallPluginCommand.getElasticUrl(InstallPluginCommand.java:292)
	at org.elasticsearch.plugins.InstallPluginCommand.download(InstallPluginCommand.java:248)
	at org.elasticsearch.plugins.InstallPluginCommand.execute(InstallPluginCommand.java:223)
	at org.elasticsearch.plugins.InstallPluginCommand.execute(InstallPluginCommand.java:210)
	at org.elasticsearch.cli.EnvironmentAwareCommand.execute(EnvironmentAwareCommand.java:86)
	at org.elasticsearch.cli.Command.mainWithoutErrorHandling(Command.java:124)
	at org.elasticsearch.cli.MultiCommand.execute(MultiCommand.java:79)
	at org.elasticsearch.cli.Command.mainWithoutErrorHandling(Command.java:124)
	at org.elasticsearch.cli.Command.main(Command.java:90)
	at org.elasticsearch.plugins.PluginCli.main(PluginCli.java:48)
Caused by: java.security.ProviderException: Could not initialize NSS
	at sun.security.pkcs11.SunPKCS11.<init>(SunPKCS11.java:223)
	at sun.security.pkcs11.SunPKCS11.<init>(SunPKCS11.java:103)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at sun.security.jca.ProviderConfig$2.run(ProviderConfig.java:224)
	at sun.security.jca.ProviderConfig$2.run(ProviderConfig.java:206)
	at java.security.AccessController.doPrivileged(Native Method)
	at sun.security.jca.ProviderConfig.doLoadProvider(ProviderConfig.java:206)
	at sun.security.jca.ProviderConfig.getProvider(ProviderConfig.java:187)
	at sun.security.jca.ProviderList.getProvider(ProviderList.java:233)
	at sun.security.jca.ProviderList.getIndex(ProviderList.java:263)
	at sun.security.jca.ProviderList.getProviderConfig(ProviderList.java:247)
	at sun.security.jca.ProviderList.getProvider(ProviderList.java:253)
	at java.security.Security.getProvider(Security.java:503)
	at sun.security.ssl.SignatureAndHashAlgorithm.<clinit>(SignatureAndHashAlgorithm.java:415)
	... 28 more
Caused by: java.io.FileNotFoundException: /usr/lib/libnss3.so
	at sun.security.pkcs11.Secmod.initialize(Secmod.java:193)
	at sun.security.pkcs11.SunPKCS11.<init>(SunPKCS11.java:218)
	... 44 more
ERROR: Service 'elasticsearch' failed to build: The command '/bin/sh -c bin/elasticsearch-plugin install --batch --silent ingest-attachment' returned a non-zero code: 1
```

docker/docker-compose.yml
nathanielks: Got it, thanks for that!

roborourke: That constant has been requested a fair bit on S3 Uploads, we may as well support it natively at this point.

roborourke:
@nathanielks does this mean #1 and #2 are already supported?

nathanielks:
@roborourke it depends on how supported you mean by "supported." The `docker-compose.yml` is set up so that `docker-compose` will assign a random port on the host (macOS) so that it's accessible there. That's retrievable by `docker-compose port db 3306` or `docker-compose port elasticsearch 9200` which will print the ip and port the service is bound to. IIRC Joe said `composer local-server` will pass through to `docker-compose`, so I suppose `composer local-server port db 3306` would print it out?

nathanielks:
@roborourke aye, and thankfully they are! In the long run, I would like to have the pattern you described, though. I think that's rather intuitive/expected behavior. Unfortunately it's a rather complicated solution. We _could_ bind to those ports on the host, but then we wouldn't be able to run multiple instances of the local server in parallel because of port conflicts. The answer would be to have some sort of proxy service that's able to map `project.example.dev:3306` to the right container, but I haven't figured out how to do that yet (it's more complicated than it sounds). It might mean introducing Consul or another service-discovery tool to the mix, but I havne't taken the time to investigate that further yet.

roborourke:
@nathanielks for the external db and es access we could use the same pattern as the s3 container or the mailhog one I just added. So `elasticsearch-<project>.altis.dev`, `db-<project>.altis.dev` etc. Or `<project>.altis.dev/elasticsearch` and `<project>.altis.dev/db`. What do you think?

nathanielks:
I personally prefer subdomains for resource addressing! 🙂 . is `{service}-(project}.altis.dev required, or can it be `{service}.{project}.altis.dev`? Ahhh, does the latter interfere with the wildcard subdomain set up on the nginx container?

nathanielks:
I wonder if `priority=1` on nginx will now make it work?

roborourke:
@nathanielks sub-sub domains would be preferable but they won't really work right now as the SSL cert is only for `*.altis.dev`.

nathanielks:
Understood. We can always update to use sub-sub domains later!

nathanielks:
@roborourke 	this is looking good! Likely unrelated, but I had some issues getting elasticsearch started. It required 2 invocations of `composer local-server start` because:

```
Warning: Error in ElasticPress request: cURL error 52: Empty reply from server (http_request_failed) in /usr/src/app/packages/enhanced-search/inc/namespace.php on line 139
Warning: Error in ElasticPress request: cURL error 7: Failed to connect to elasticsearch port 9200: Connection refused (http_request_failed) in /usr/src/app/packages/enhanced-search/inc/namespace.php on line 139
```

The second time round it was able to boot proper. I tested the urls and things work as desired! I had expected `altis-dev.altis.dev/uploads` to have worked for the uploads path like Tachyon, but I realized `s3-altis-dev.altis.dev/uploads` was the way to do it. Is it possible to also have it be on the same sub-domain as tachyon to replicate production behavior?

nathanielks:
I made this modification to test and it doesn't appear so:
Screen Shot 2019-07-01 at 7 35 31 AM

I defer to you though, you may know more on this than me at this point!

roborourke:
@nathanielks I think the bucket URL would need to end with `/uploads` too, but I'm not sure if that should be `/uploads` or just something like `<project>.altis.dev/s3`

Weird about the elasticsearch error. I'll try again with a clean install.

nathanielks:
Definitely `/uploads` to mirror the cloud environment 😃

joehoyle:
@roborourke it seems ES got quite a bit slower with this change. Maybe the mem limit changes? I've had to increase the healthcheck interval, but ideally we wouldn't have to do that. Do you know why we needed to change the mem limit? I'm assuming before it had more memory so started up faster.

roborourke:
@joehoyle we could maybe increase it. Basically docket machine has a default limit of 2Gb. If you spun up a second instance the other Elasticsearch instance was first for the chopping block without any limits set.

Can always recommend folks only run one stack at a time I guess.

joehoyle:
Ah hmm, I _think_ it would be ok to get people to increase the docker engine ram, I guess I had put it to 8GB at some point:



ES is totes a memory hog though!

roborourke:
Will def need to document that anyway. Maybe we can double what I put in this PR. Should be fine for local dev

nathanielks:
> Can always recommend folks only run one stack at a time I guess.

Nah, I wouldn't want that. I'd just document this and make note that if you _want_ to run more than one environment, be sure to increase Docker's available ram.

roborourke:
@nathanielks re. the S3 URL, seems none of the local S3 options really support serving files from a subdirectory, certainly not fakes3, minio or riak-cs so we're stuck with the subdomain on local :(

nathanielks:
Got it! A-okay, it can always be documented till we figure out a way around it!

PR #10, opened 2019-04-08
Add mock PHPMailer handler

joehoyle:
This sends all emails to the PHP log. Better than nothing, but an `email` logs / microserver would be better i think


rmccue:
Similar to Chassis, you could consider [MailHog](https://github.com/mailhog/MailHog) here?

roborourke:
Agree something like mailhog will make for a better dev experience here

PR #11, opened 2019-04-12
Add restart and shell commands

joehoyle:
Restart for convenience, shell to open a bash shell on the php container.

### DIFF
--- a/inc/Composer/Command.php
+++ b/inc/Composer/Command.php
@@ -47,12 +47,16 @@ EOT
 			return $this->start( $input, $output );
 		} elseif ( $subcommand === 'stop' ) {
 			return $this->stop( $input, $output );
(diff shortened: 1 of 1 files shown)


inc/Composer/Command.php
```diff
@@ -171,8 +184,20 @@ protected function logs( InputInterface $input, OutputInterface $output ) {
 			'VOLUME' => getcwd(),
 			'COMPOSE_PROJECT_NAME' => basename( getcwd() ),
 		] );
+		$compose->setTimeout( 0 );
```
roborourke: curious what this is for?

joehoyle: When tailing logs, the command is kept running, so we don't want to abort after 60 seconds

PR #12, opened 2019-04-15
Add check for add_action

joehoyle:
Per https://github.com/humanmade/platform-dev/issues/323, we are screwing up any inclusion of autoload.php outside of the wp-config.php. We are discussing a better long term solution in https://github.com/humanmade/platform-dev/pull/325, but this is intended to get things functioning.


rmccue:
@joehoyle You've mixed in a `define` change here too?

joehoyle:
Ah my bad

PR #13, opened 2019-04-17
Remove trailing comma on function call.

peterwilsoncc:


joehoyle:
Wow, is that a fatal? Unsure how this snuck in, unless it's cmpatible with 7.3 or something?

peterwilsoncc:
Yeah, it's PHP 7.3+ only.

rmccue:
Once we go 7.3 minimum, looking forward to adding this to our coding standards 🙏

PR #17, opened 2019-05-09
Makes the tachyon service use a path

### LINKED ISSUE #16: CORS headers for tachyon
Right now the tachyon service is using a different domain name to deliver the images. There are no CORS headers that allow loading of those images via javascript which breaks the smart media cropping.

Possible options are:

- Add more permissive CORS headers
- Use a sub-sub domain eg. *.*.altis.dev
- Use a sub folder path and proxy_pass via nginx

roborourke:
This matches our hosting stacks and fixes #16

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -13,8 +13,6 @@ x-php: &php
   links:
     - "db:db-read-replica"
     - "s3:s3-${COMPOSE_PROJECT_NAME:-default}.s3.localhost"
-  external_links:
-    - "proxy:tachyon.${COMPOSE_PROJECT_NAME:-default}.altis.dev"
   volumes:
     - "${VOLUME}:/usr/src/app:delegated"
   environment:
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -14,7 +14,7 @@ x-php: &php
     - "db:db-read-replica"
     - "s3:s3-${COMPOSE_PROJECT_NAME:-default}.s3.localhost"
   external_links:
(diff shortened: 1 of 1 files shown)


docker/docker-compose.yml
```diff
@@ -14,7 +14,7 @@ x-php: &php
     - "db:db-read-replica"
     - "s3:s3-${COMPOSE_PROJECT_NAME:-default}.s3.localhost"
   external_links:
-    - "proxy:tachyon-${COMPOSE_PROJECT_NAME:-default}.altis.dev"
+    - "proxy:tachyon.${COMPOSE_PROJECT_NAME:-default}.altis.dev"
```
roborourke: Can probably remove this bit altogether

docker/docker-compose.yml
```diff
@@ -136,7 +134,7 @@ services:
       - "traefik.port=8080"
       - "traefik.protocol=http"
       - "traefik.docker.network=proxy"
-      - "traefik.frontend.rule=HostRegexp:tachyon-${COMPOSE_PROJECT_NAME:-default}.altis.dev"
+      - "traefik.frontend.rule=HostRegexp:${COMPOSE_PROJECT_NAME:-default}.altis.dev,{subdomain:[a-z.-_]+}.${COMPOSE_PROJECT_NAME:-default}.altis.dev;PathPrefix:/tachyon;ReplacePathRegex:^/tachyon/(.*) /uploads/$$1"
```
nathanielks: For local, does this need to be `/content/uploads`?

roborourke: No this is working because of the S3 Uploads rewrite as far as I can tell

PR #29, opened 2019-05-17
Add healthcheck to Elasticsearch

### LINKED ISSUE #27: Error on install with elasticsearch
I think this when installing for the first time:

> Warning: Error in ElasticPress request: cURL error 7: Failed to connect to elasticsearch port 9200: Connection refused (http_request_failed) in /usr/src/app/vendor/altis/enhanced-search/inc/namespace.php on line 127

joehoyle:
Fixes #27.

### DIFF
--- a/docker/elasticsearch/Dockerfile
+++ b/docker/elasticsearch/Dockerfile
@@ -1,2 +1,3 @@
 FROM blacktop/elasticsearch:6.3
 RUN bin/elasticsearch-plugin install --batch --silent ingest-attachment
+RUN apk add --no-cache curl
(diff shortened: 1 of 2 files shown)


docker/elasticsearch/Dockerfile
```diff
@@ -1,2 +1,3 @@
 FROM blacktop/elasticsearch:6.3
 RUN bin/elasticsearch-plugin install --batch --silent ingest-attachment
+RUN apk add --no-cache curl
```
roborourke: Take you've pushed the latest image? We should split this out into its own repo so releases are autobuilt & published

joehoyle: Yup pushed

PR #30, opened 2019-05-17
Add loopback http support

joehoyle:
Currently you can't make internal requests to the site url, as `*.altis.dev` resolves to 127.0.0.1 which is the IP of the container form internal requests. Instead we want to link the hostname to the traefik container.


joehoyle:
Yeah the cert is only valid for sub domains.

PR #33, opened 2019-05-24
Fix coding standards

roborourke:


roborourke:
@joehoyle using psr-4 is making it hard to get this compatible with HM Linter. I'm not sure why the `phpcs:disable` lines are working for the HM file naming rules but not the WordPress ones...

rmccue:
I think we should switch from PSR-4 to our standards for file naming, and use the classmap autoloader.

joehoyle:
I'm not sure you can use ClassMap for plugins, there's only some of the autoloading methods that are supported for plugins I think.

roborourke:
@joehoyle `altis/core` uses classmap https://github.com/humanmade/altis-core/blob/master/composer.json

joehoyle:
👍

inc/namespace.php
```diff
@@ -0,0 +1,68 @@
...
+		 * In local-server, the tachyon hostname resolves to what is deemed a local url.
+		 * This makes requests to tachyon from WordPress disallowed. We want to
+		 * specifically allow that host.
+		 */
+		add_filter( 'http_request_host_is_external', function ( bool $is_external, string $host ) : bool {
+			if ( $is_external ) {
+				return $is_external;
+			}
+
+			// @codingStandardsIgnoreLine
```
rmccue: What is this ignoring? 🤔

roborourke: `parse_url()` vs `wp_parse_url()`

PR #37, opened 2019-06-11
Add mailhog service

### LINKED ISSUE #36: Add a mock email handler
[Mailhog](github.com/mailhog/mailhog) is well established tool for this job.

roborourke:
Supercedes #10
Fixes #36

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -21,6 +21,7 @@ x-php: &php
     - "proxy:${COMPOSE_PROJECT_NAME}.altis.dev"
   volumes:
     - "${VOLUME}:/usr/src/app:delegated"
+    - "${PWD}/php.ini:/usr/local/etc/php/conf.d/altis.ini"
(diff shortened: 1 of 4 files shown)


roborourke:
@nathanielks the frontend rules are matched in order of their length by default so because the nginx one is kind of a catch all for subdomains it means the `elasticsearch` subdomain was never working externally and it incorrectly had a match for all sub-sub domains eg. *.<project>.altis.dev

Because I was running into issues using sub-subdomains for services redirecting to https it was easier to place at them on subdirectories, but again because those were shorter than the nginx frontend rule that would kick in. This way the more specific but shorter frontend rules are matched before falling back to nginx as a catch all.

roborourke:
@nathanielks yeah actually, could set it to have a priority of 1 which is equivalent to it being 1 character long. 0 means use the full length. I'll make that update now.

nathanielks:
Perfect!

PR #38, opened 2019-06-28
Local analytics support

roborourke:
Adds a container to work with and test analytics data locally by mocking cognito and pinpoint.

Also adds Kibana for debugging and exploring analytics data.


docker/docker-compose.yml
```diff
@@ -173,6 +175,36 @@ services:
...
+      - proxy
+      - default
+    labels:
+      - "traefik.port=3000"
+      - "traefik.protocol=http"
+      - "traefik.docker.network=proxy"
+      - "traefik.frontend.rule=HostRegexp:analytics-${COMPOSE_PROJECT_NAME:-default}.altis.dev"
+    restart: on-failure
+  kibana:
+    image: blacktop/kibana:6.3
```
joehoyle: Is this an official image, could we get / use the elastic one here?

roborourke: It's the OSS version of KIbana. We're using `blacktop/elasticsearch` for the ES service so didn't think it was an issue. To use the official image requires building an image that uninstalls x-pack which this one doesn't have.

joehoyle:
Just getting this set up @roborourke, while not strictly for local-server, thought I'd ask here anyway, as we don't (I think) have the main PR up for this feature.

I've got my stuff running locally, and I also had to add

```php
define( 'HM_ANALYTICS_PINPOINT_ID', '1' );
define( 'HM_ANALYTICS_PINPOINT_REGION', 'us-east-1' );
define( 'HM_ANALYTICS_COGNITO_ID', '1' );
define( 'HM_ANALYTICS_COGNITO_REGION', 'us-east-1' );
```

To my config. I now have an index called `analytics` too in my es cluster, however I've not getting any docs in the index. I can see the front-end us indeed sending requests to pinpoint (though weirdly that url with the `...` looks a bit suspect:



And the response:



Which I guess all looks good.

Given I (think) there's no firehose locally, how can I see / debug why pinpoint doesn't appear to be adding the events to the ES cluster?

joehoyle:
Doh! As I posted that I now see docs! I guess pinpoint is batching it perhaps

roborourke:
So the `...` would normally be the pinpoint app ID.

For debugging you can check the output of each container via the logs command. The local cognito and pinpoint images are at github.com/humanmade/local-cognito and github.com/humanmade/local-pinpoint if you want to edit and rebuild them locally. 

If the index is there then it’s partly working but perhaps the SDK needs more legit looking values... I’m not sure on that tbh.

roborourke:
It needs splitting out some but check out the record and flushEvents functions in tracker.js of the analytics plugin. Not made the PR for that yet.

roborourke:
@joehoyle good to merge?

PR #40, opened 2019-07-05
Allow project subdomain to be configured.

### LINKED ISSUE #35: Hostname fails if current dir has a .
If the current DIR is not a valid subdomain, things fail. E.g. `example.com.altis.dev`. We should sanitize this.

joehoyle:
This will also sanitize the name to be subdomain friedly.

Fixes #35.

### DIFF
--- a/inc/composer/class-command.php
+++ b/inc/composer/class-command.php
@@ -74,7 +74,7 @@ EOT
 
 		$compose = new Process( 'docker-compose up -d', 'vendor/altis/local-server/docker', [
 			'VOLUME' => getcwd(),
-			'COMPOSE_PROJECT_NAME' => basename( getcwd() ),
+			'COMPOSE_PROJECT_NAME' => $this->get_project_subdomain(),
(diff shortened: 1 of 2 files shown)


inc/composer/class-command.php
```diff
@@ -232,9 +232,27 @@ protected function shell( InputInterface $input, OutputInterface $output ) {
...
+	/**
+	 * Get the name of the project for the local subdomain
+	 *
+	 * @return string
+	 */
+	protected function get_project_subdomain() : string {
+
+		$composer_json = json_decode( file_get_contents( getcwd() . '/composer.json' ), true );
+
+		if ( isset( $composer_json['extra']['altis']['modules']['local-server']['name'] ) ) {
```
roborourke: I thought you had to check each stage of the array tree when using `isset()` so I've used null coalescing in most cases like this. If there're no warnings then TIL!

rmccue: Nope! https://3v4l.org/9e8R5

PR #56, opened 2019-10-09
Map QM paths to host file paths

tomjn:
Tells qm how to adjust paths in stack traces to use the host machines file paths

See https://github.com/humanmade/altis-dev-tools/issues/16


inc/namespace.php
```diff
@@ -75,4 +75,32 @@ function bootstrap() {
...
+}
+
+/**
+ * Enables Query Monitor to map paths to their
+ * original values on the host.
+ *
+ * @param array $map an associative array of folders as keys, and their mappings as values
+ * @return array the adjusted mapping of folders
+ */
+function set_file_path_map( $map ) : array {
```
pdewouters: `array $map`?

tomjn: agreed

inc/namespace.php
```diff
@@ -75,4 +75,32 @@ function bootstrap() {
...
+function set_file_path_map( array $map ) : array {
+	if ( ! file_exists( '/etc/chassis-constants' ) ) {
+		return $map;
+	}
+	$json_string = file_get_contents( '/etc/chassis-constants' );
+	if ( empty( $json_string ) ) {
+		return $map;
+	}
+	$data = json_decode( $json_string, true );
+	if ( empty( $data['synced_folders'] ) ) {
```
rmccue: I'd prefer an explicit `empty( $data )` here, as it's a different code path.
```suggestion
	if ( empty( $data ) || empty( $data['synced_folders'] ) ) {
```

inc/namespace.php
```diff
@@ -75,4 +75,32 @@ function bootstrap() {
...
+
+	add_filter( 'qm/output/file_path_map', __NAMESPACE__ . '\\set_file_path_map', 1 );
+}
+
+/**
+ * Enables Query Monitor to map paths to their
+ * original values on the host.
+ *
+ * @param array $map an associative array of folders as keys, and their mappings as values
+ * @return array the adjusted mapping of folders
```
rmccue: Comments should be updated per https://github.com/humanmade/altis-local-chassis/pull/39#pullrequestreview-300798375

tomjn:
This should be fine, but, for whatever reason `composer` doesn't pick up the `local-chassis` subcommand with this branch on my local instance, likely due to the namespace change, but I'm not seeing what the culprit is 🤔

tomjn:
`composer local-server` is working locally now, false alarm :/

inc/namespace.php
```diff
@@ -75,4 +75,31 @@ function bootstrap() {
...
+}
+
+/**
+ * Enables Query Monitor to map paths to their original values on the host.
+ *
+ * @param array $map Map of guest path => host path
+ * @return array Adjusted mapping of folders
+ */
+function set_file_path_map( array $map ) : array {
+	if ( ! file_exists( '/etc/chassis-constants' ) ) {
```
roborourke: Is this meant to be in the local chassis repo? This file won't exist on local server

roborourke: Odd. I figured out an alternative method that works at least. Cheers for getting this off the ground. It’s very cool

PR #64, opened 2019-10-21
XDebug/dev php server support

roborourke:
This update allows passing an `--xdebug` flag that starts the main PHP server with XDebug activated. Having it always on is not ideal in all situations as PHP runs considerably slower.

Still to do:

- [x] Add Docs


docs/troubleshooting.md
```diff
@@ -22,7 +22,7 @@ All containers should have a status of "Up". If they do not, you can inspect the
 
 ## Services keep stopping
 
-By default docker machine sets a default memory limit of 2gb for all of your containers. Because of this if your system becomes too busy or you're running multiple instances of local server it is recommended to increase this limit to at least 4gb.
+By default docker machine sets a default memory limit of 2Gb for all of your containers. Because of this if your system becomes too busy or you're running multiple instances of local server it is recommended to increase this limit to at least 4Gb.
```
rmccue:
```suggestion
By default docker machine sets a default memory limit of 2GB for all of your containers. Because of this if your system becomes too busy or you're running multiple instances of local server it is recommended to increase this limit to at least 4GB.
```

PR #71, opened 2019-10-31
Ensure S3 bucket and uploads dir exist on start

roborourke:
WordPress and in particular WP unit tests don't factor in a missing uploads directory. Because we're using fakes3 the bucket and directories aren't created until an image is first uploaded. This will avoid errors being thrown.

### DIFF
--- a/inc/composer/class-command.php
+++ b/inc/composer/class-command.php
@@ -143,6 +143,15 @@ EOT
 				'--quiet',
 			],
 		] ), $output );
+		$cli->run( new ArrayInput( [
+			'subcommand' => 'cli',
+			'options' => [
+				's3-uploads',
+				'rm',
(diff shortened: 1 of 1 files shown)


inc/composer/class-command.php
```diff
@@ -129,6 +131,19 @@ protected function start( InputInterface $input, OutputInterface $output ) {
 			$output->writeln( '<info>WP Password:</>	<comment>admin</>' );
 		}
 
+		// Ensure uploads directory is present by copying a known file.
+		// Prevents errors when running WordPress unit tests.
+		$cli->run( new ArrayInput( [
```
joehoyle: Can we delete this file after it's done?

roborourke: @joehoyle yep, updated

PR #72, opened 2019-11-01
Add exec command

roborourke:
This modifies the existing cli command to be a generic exec command allowing for running things like `vendor/bin/phpunit`.

The exec method also accepts a default program to pass the options onto.


inc/composer/class-command.php
```diff
@@ -36,6 +36,8 @@ protected function configure() {
 	status
 Run WP CLI command:
 	cli -- <command>              eg: cli -- post list --debug
+Run any command:
```
joehoyle:
```suggestion
Run any shell command from the PHP container:
```

PR #81, opened 2019-11-22
Restart cavalcade on failure

roborourke:
Because the server is ready before WP is installed the runner starts and finds the table doesn't exist then promptly exits and stops the container. Not ideal!

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -116,6 +116,7 @@ services:
     entrypoint:
       - /usr/local/bin/cavalcade
     user: "nobody:nobody"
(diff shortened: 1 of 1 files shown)


roborourke:
@rmccue that's not an available option, but presumably if there's some PHP error or whatever it should come back up. Guessing we do that on our infra?

rmccue:
On the infra, I believe it will retry ~5 times; it also retries again (5 times) when you deploy.

roborourke:
I'll see what we can do in future but right now I can't find another way to ensure the service is available after the initial install.

PR #89, opened 2019-12-02
Add shared socket volume for nginx/php socket communication

nathanielks:
https://github.com/humanmade/docker-wordpress-php/pull/28 and https://github.com/humanmade/docker-wordpress-nginx/pull/13 update nginx/php communication to use a Unix Socket instead of TCP. This adds a shared volume for them to communicate over. Once those two PRs are merged and new tags released, we can update the tags referenced in `docker-compose.yml`.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -22,6 +22,7 @@ x-php: &php
   volumes:
     - "${VOLUME}:/usr/src/app:delegated"
     - "${PWD}/php.ini:/usr/local/etc/php/conf.d/altis.ini"
+    - socket:/var/run/php-fpm
   environment:
     HOST_PATH: ${VOLUME}
     DB_HOST: db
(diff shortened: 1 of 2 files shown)


joehoyle:
Looks good, let's wait until those are tagged and then we can update this PR with all changes at once

nathanielks:
Versions updated. Docker Hub builds are still in progress.

nathanielks:
Builds are complete!

docker/docker-compose.yml
```diff
@@ -83,18 +84,19 @@ services:
       - '6379'
   php:
     <<: *php
-    image: ${PHP_IMAGE:-humanmade/altis-local-server-php:3.0.0}
+    image: ${PHP_IMAGE:-humanmade/altis-local-server-php:3.1.0}
```
roborourke: Need to update this for the `start` command too: https://github.com/humanmade/altis-local-server/blob/master/inc/composer/class-command.php#L103

nathanielks: Thanks, Rob! Updated 👍

roborourke:
Shall we backport for v3?

nathanielks:
@roborourke label added 👍 . I see now... merges into `master` do not get merged into any specific branch unless the label is explicitly added?

PR #91, opened 2019-12-04
Change command name to `server`

### LINKED ISSUE #87: Alias "local-server" command to "server"
Writing `composer local-server <command>` is needlessly long and the hyphen isn't great for quick tab completion.

Let's do two things:

- add an alias so we can write `composer server start`
- make `start` the default command so it's just `composer server`

roborourke:
Fixes #87

### DIFF
--- a/inc/composer/class-command.php
+++ b/inc/composer/class-command.php
@@ -74,10 +74,13 @@ EOT
 			return $this->logs( $input, $output );
 		} elseif ( $subcommand === 'shell' ) {
 			return $this->shell( $input, $output );
+		} elseif ( $subcommand === '' ) {
+			// Default to start command.
+			return $this->start( $input, $output );
 		}
 
-		// Default to start command.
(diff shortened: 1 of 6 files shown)


inc/composer/class-command.php
```diff
@@ -74,8 +76,8 @@ protected function execute( InputInterface $input, OutputInterface $output ) {
 			return $this->shell( $input, $output );
 		}
 
-		$output->writeln( '<error>' . $subcommand . ' command not found.</>' );
-		return 1;
+		// Default to start command.
+		return $this->start( $input, $output );
```
rmccue: I think this should check if subcommand is empty, otherwise something like `composer serve foo` will trigger this.

roborourke: Good call, updated

nathanielks:
Nice, I like it 👍

PR #116, opened 2020-01-24
Don't send empty cookie on every request

joehoyle:
When not using `PHP_XDEBUG_ENABLED`, we currently send a blank cookie on _every_ request. This is wasteful, but also means you won't get Batcache locally because of cookies being set, and can snag other problems too.

I don't really see why we even need to set this at all. In this context.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -79,8 +79,6 @@ function bootstrap() {
 	// Set XDebug cookie if environment variable is set.
 	if ( getenv( 'PHP_XDEBUG_ENABLED' ) ) {
 		setcookie( 'XDEBUG_SESSION', $_SERVER['HTTP_HOST'], strtotime( '+1 year' ), '/', $_SERVER['HTTP_HOST'] );
-	} else {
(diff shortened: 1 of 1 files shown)


roborourke:
This is to unset the cookie if previously set when you start without xdebug, so I’d say detect if it’s actually set first then unset it in that scenario. The reason for doing this is so you don’t need one of those browser addons to enable / disable xdebug logging when you start the server with xdebug. Basically avoids an additional manual step.

joehoyle:
If the cookie is set, but you don’t have code if enabled, why unset it? Seems fine to just stay there?

-- 

Joe Hoyle

CTO
Human Made

roborourke:
Good point 🤦‍♂

roborourke:
Now wondering if we should check whether the `XDEBUG_SESSION` cookie is set before sending it with a new expiry every time when you do have xdebug enabled.

joehoyle:
Mmm maybe. Sounds like a follow-up issue / PR :D

PR #119, opened 2020-01-29
Add Windows support

joehoyle:
Currently many things are broken in Windows, due to not testing paths, env vars, ttys and more. This fixes up all the errors I ran into, until I get a functioning Windows setup!

This was opened in the GH web UI, from 3.* so this diff needs cleaning still (e.g. ` 				$output->writeln( sprintf( '<error>WordPress install failed. Exited with error code %d</>', $install_failed ) );` should not have been deleted.

### DIFF
--- a/inc/composer/class-command.php
+++ b/inc/composer/class-command.php
@@ -86,9 +86,18 @@ EOT
 	protected function start( InputInterface $input, OutputInterface $output ) {
 		$output->writeln( '<info>Starting...</>' );
 
-		$proxy = new Process( 'docker-compose -f proxy.yml up -d', 'vendor/altis/local-server/docker' );
+		// Don't set the composer project env var for the proxy, as it's
+		// intended to be shared for all projects.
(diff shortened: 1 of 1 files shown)


joehoyle:
Also, we need to add support for testing of `$_ENV` is empty. On windows, the `php.ini` defaults to GPCS (per https://www.php.net/manual/en/ini.core.php#ini.variables-order) so we should add a check for that.

roborourke:
Nice! Can that ini setting be changed at run time with `ini_set()` or setting `auto_globals_jit`?

Code looks good apart from the linter errors.

joehoyle:
I think no, as ENV is populated on the process bootstrap

-- 

Joe Hoyle

CTO
Human Made

rmccue:
@joehoyle Given we have WSL2 support, is there anything else in this PR that might be worth retaining?

roborourke:
🔥

joehoyle:
Yeah can be deleted!

PR #120, opened 2020-02-04
Added Kibana and MailHog links in WP Admin

### LINKED ISSUE #106: Add links to kibana and mailhog to WP admin tools menu
We should make it easier to find mailhog and kibana by linking to them from the tools menu in WP admin.

igmoweb:
Captura de pantalla 2020-02-04 a las 11 05 50

Fixes #106

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -11,8 +11,6 @@ use function Altis\get_environment_type;
 function bootstrap() {
 	if ( get_environment_type() === 'local' ) {
 		add_filter( 'admin_menu', function() {
-			global $submenu;
-
 			$links = [
 				[
 					'label' => 'Kibana',
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -3,15 +3,12 @@
 namespace Altis\Local_Server;
 
 use function Altis\get_config;
-use function Altis\get_environment_architecture;
 
 /**
  * Configure environment for local server.
  */
 function bootstrap() {
-	if ( get_environment_architecture() === 'local-server' ) {
-		add_filter( 'admin_menu', __NAMESPACE__ . '\\tools_submenus' );
-	}
+	add_filter( 'admin_menu', __NAMESPACE__ . '\\tools_submenus' );
 
 	$config = get_config()['modules']['local-server'];
 
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -3,11 +3,33 @@
 namespace Altis\Local_Server;
 
 use function Altis\get_config;
+use function Altis\get_environment_type;
 
 /**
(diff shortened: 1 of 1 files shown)


inc/namespace.php
```diff
@@ -3,11 +3,31 @@
...
 
 use function Altis\get_config;
+use function Altis\get_environment_type;
 
 /**
  * Configure environment for local server.
  */
 function bootstrap() {
+	if ( get_environment_type() === 'local' ) {
+		add_filter( 'admin_menu', function() {
```
roborourke: I know it's not all that consistent where we use anonymous functions in this file but I think it'll keep things tidier from now on to just have the filter call here.

inc/namespace.php
```diff
@@ -3,11 +3,31 @@
 namespace Altis\Local_Server;
 
 use function Altis\get_config;
+use function Altis\get_environment_type;
 
 /**
  * Configure environment for local server.
  */
 function bootstrap() {
+	if ( get_environment_type() === 'local' ) {
```
roborourke: This code is only run if local server is in use so you don't need this check here, see the `wp-config.php` and `load.php` to see how that's achieved:

https://github.com/humanmade/altis-local-server/blob/master/load.php

inc/namespace.php
roborourke: Sorry to clarify this I meant:

- move the anonymous function out to the bottom of the file
- call `add_filter( 'admin_menu', __NAMESPACE__ . '\\tools_menu' );` or whatever you name the function

igmoweb:
@roborourke Ready to be reviewed again.

inc/namespace.php
```diff
@@ -3,11 +3,16 @@
 namespace Altis\Local_Server;
 
 use function Altis\get_config;
+use function Altis\get_environment_architecture;
 
 /**
  * Configure environment for local server.
  */
 function bootstrap() {
+	if ( get_environment_architecture() === 'local-server' ) {
```
roborourke: Sorry for the back and forth, I wasn't super clear again. This check is already done here:

https://github.com/humanmade/altis-local-server/blob/master/load.php#L10

When a module is registered the `bootstrap` callback is only run if the `enabled` setting is true, so you don't need to check for this again here.

If someone was to manually enable the module in a different context for some reason then presumably they know what they're doing. If we didn't trust them to know what they're doing we'd be better off making this check into a general way to return early eg:

```php
if ( get_environment_architecture() === 'local-server' ) {
     return;
}
```

The rest of the code in the `bootstrap()` function assumes you're running within the docker environment - eg. looking at `getenv()` to set constants and so on.

roborourke:
Nice one 👍

PR #121, opened 2020-02-04
Escape WP CLI queries

igmoweb:
When running things like `composer server cli -- db query 'select * from wp_posts;'`, the query was not taken as a single argument. Identifying the query position and escaping it solved it


roborourke:
Not sure it was me but I was wondering if there's not an easier way - this method seemed specific to the example use case eg. passing a db query via the proxied CLI command. I was wondering if it was possible to take the whole `$options` list, implode it, wrap in quotes as a single argument and escape it.

igmoweb:
@rmccue @roborourke I've simplified the `foreach` loop (I couldn't swap it into an `implode` as there's a case that needs some string transformation). Could you think on any edge case that could not work with the change?

rmccue:
@igmoweb I think the concern is more the level it's escaping at; we shouldn't need to parse the string to find `=` or anything, and that's probably an indication that our escaping later on is wrong.

I _think_ using `escapeshellarg` when we sprintf `$command` together is where we should be escaping it.

igmoweb:
@rmccue I think that every argument should be escaped separate, otherwise, the command will end up like:

`docker exec ... wp 'db query select * from wp_posts; --url=https://altis-test.altis.dev/'`

And that doesn't work, however this one does (this is how it currently works):

`docker exec ... wp 'db' 'query' 'select * from wp_posts;' '--url=https://altis-test.altis.dev/'`

So at least we should map through every argument in options list and escape them.

Also, I kept the `=` to keep backward compatibility, it was weird to see but I didn't want to remove it just in case.

I've pushed another amplification, does it look better?

rmccue:
> otherwise, the command will end up like

Actually, I think that is basically what we want. When I run:

```sh
composer server cli -- foo bar 'quux'
```

Under the hood it should run
```
docker exec ... 'wp foo bar \'quux\''
```

In your above example, it looks like we're mistakenly stripping quotes from the arguments, which is the first problem. (That may be a Symfony console issue.) If we can fix that, then the escapeshellarg later should work fine (although `wp` needs to be inside that string too)

igmoweb:
@rmccue I'm running out of ideas here. I just pushed another commit to leave the command working, however, is not properly escaped.

A command like this
`composer server cli -- search-replace "Hello world" "whatever2" --url=https://altis-test.altis.dev/ --dry-run`

would run a Docker command  like the following one:

`docker exec -e COLUMNS=134 -e LINES=62 -u www-data  1672111c8e61 wp 'search-replace' 'Hello world' 'whatever2' '--url=https://altis-test.altis.dev/' '--dry-run'`

It works but it's not ideal as command options are escaped one by one instead of escaping the whole `wp` command.

Any ideas?

roborourke:
The problem no longer seems to occur so closing this out. Will wait and see if this comes again in future.

PR #122, opened 2020-02-06
Add Database Connection Helper Command

igmoweb:
WIP

Issue: #1

- [x] `composer server db`
- [x] `composer server db info`
- [x] `composer server db spf`

`composer server db info` sample:
Captura de pantalla 2020-02-06 a las 16 23 59

### DIFF
--- a/docs/database.md
+++ b/docs/database.md
@@ -6,7 +6,7 @@ You can log into MySQL and run any query in the default database (`wordpress`) b
 composer server db
 ```
 
-Use `composer server info` to retrieve MySQL info, connection details and forwarded ports:
+Use `composer server db info` to retrieve MySQL info, connection details and forwarded ports:
 
 ```sh
 Root password:  wordpress
@@ -16,6 +16,8 @@ User:           wordpress
 Password:       wordpress
 
 Version:        5.7.26-1debian9
-Ports:          0.0.0.0:32791
+Host:           0.0.0.0
+Port:           32775
 ```
 
+Use `composer server db sequel` to generate an SPF file. This command can only be run under MacOS and requires [Sequel Pro](https://www.sequelpro.com/) to be installed in your computer. 
--- a/docs/database.md
+++ b/docs/database.md
@@ -6,7 +6,7 @@ You can log into MySQL and run any query in the default database (`wordpress`) b
 composer server db
 ```
 
-Use `composer server db info` to retrieve MySQL info, connection details and forwarded ports:
+Use `composer server db info` to retrieve MySQL info and connection details:
 
 ```sh
 Root password:  wordpress
@@ -15,9 +15,10 @@ Database:       wordpress
 User:           wordpress
 Password:       wordpress
 
-Version:        5.7.26-1debian9
 Host:           0.0.0.0
 Port:           32775
+
+Version:        5.7.26-1debian9
 ```
 
-Use `composer server db sequel` to generate an SPF file. This command can only be run under MacOS and requires [Sequel Pro](https://www.sequelpro.com/) to be installed in your computer. 
+Use `composer server db sequel` to open the database in Sequel Pro. This command can only be run under MacOS and requires [Sequel Pro](https://www.sequelpro.com/) to be installed on your computer.
--- a/docs/database.md
+++ b/docs/database.md
@@ -1,12 +1,12 @@
-# Interacting with Database
+# Interacting with the Database
 
-- You can log into MySQL and run any query in the default database (`wordpress`) by using:
+You can log into MySQL and run any query in the default database (`wordpress`) by using:
 
 ```sh
 composer server db
 ```
 
-- Use `composer server info` to retrieve MySQL info, connection details and forwarded ports:
+Use `composer server info` to retrieve MySQL info, connection details and forwarded ports:
 
 ```sh
 Root password:  wordpress
--- a/docs/database.md
+++ b/docs/database.md
@@ -1,0 +1,21 @@
+# Interacting with Database
+
+- You can log into MySQL and run any query in the default database (`wordpress`) by using:
+
+```sh
+composer server db
+```
+
+- Use `composer server info` to retrieve MySQL info, connection details and forwarded ports:
+
+```sh
+Root password:  wordpress
+
+Database:       wordpress
+User:           wordpress
+Password:       wordpress
+
+Version:        5.7.26-1debian9
+Ports:          0.0.0.0:32791
+```
+
--- a/inc/composer/class-command.php
+++ b/inc/composer/class-command.php
@@ -372,42 +372,37 @@ EOT
 <info>Password</info>:       ${connection_data['MYSQL_PASSWORD']}
 
 <comment>Version</comment>:        ${connection_data['MYSQL_VERSION']}
-<comment>Host</comment>:          ${connection_data['HOST']}
-<comment>Port</comment>:          ${connection_data['PORT']}
+<comment>Host</comment>:           ${connection_data['HOST']}
+<comment>Port</comment>:           ${connection_data['PORT']}
 
 EOT;
 				$output->write( $db_info );
 				break;
 			case 'spf':
+				if ( strpos( php_uname(), 'Darwin' ) === false ) {
+					$output->writeln( '<error>This command is only supported on MacOS, use composer server db info to see the database connection details.</error>' );
+					return 1;
+				}
+
 				$connection_data = $this->get_db_connection_data();
-				$rpf_file_contents = file_get_contents( getcwd() . '/vendor/altis/local-server/templates/sequel.xml' );
+				$spf_file_contents = file_get_contents( dirname( __DIR__, 2 ) . '/templates/sequel.xml' );
 				foreach ( $connection_data as $field_name => $field_value ) {
-					$rpf_file_contents = preg_replace( "/(<%=\s)($field_name)(\s%>)/i", $field_value, $rpf_file_contents );
+					$spf_file_contents = preg_replace( "/(<%=\s)($field_name)(\s%>)/i", $field_value, $spf_file_contents );
 				}
-				$output_file_path = getcwd() . '/sequel.spf';
-				file_put_contents( $output_file_path, $rpf_file_contents );
+				$output_file_path = sprintf( '/tmp/%s.spf', $this->get_project_subdomain() );
+				file_put_contents( $output_file_path, $spf_file_contents );
 
-				$output_message = <<<EOT
-The SPF file has been placed in <info>$output_file_path</info>.
-
-EOT;
-
-				$output->write( $output_message );
-
-				$os = php_uname();
-				if ( strpos( $os, 'Darwin' ) !== false ) {
-					exec( "open $output_file_path", $null, $return_val );
-				} elseif ( strpos( $os, 'Windows' ) !== false ) {
-					exec( "start $output_file_path", $null, $return_val );
-				}
-
+				exec( "open $output_file_path", $null, $return_val );
 				if ( $return_val !== 0 ) {
 					$output->writeln( '<error>You must have Sequel Pro installed to use this command</error>' );
 				}
 
 				break;
-			default:
+			case null:
 				passthru( "$base_command mysql --database=wordpress --user=root -pwordpress", $return_val );
+				break;
+			default:
+				$output->writeln( "<error>The subcommand $db is not recognized</error>" );
 		}
 
 		return $return_val;
--- a/inc/composer/class-command.php
+++ b/inc/composer/class-command.php
@@ -339,16 +339,59 @@ EOT
 	}
 
 	protected function db( InputInterface $input, OutputInterface $output ) {
+		$db = $input->getArgument( 'options' )[0] ?? '';
 		$columns = exec( 'tput cols' );
 		$lines = exec( 'tput lines' );
-		passthru( sprintf(
-			'cd %s; VOLUME=%s COMPOSE_PROJECT_NAME=%s docker-compose exec -e COLUMNS=%d -e LINES=%d db mysql --database=wordpress --user=root -pwordpress',
+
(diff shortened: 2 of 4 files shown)


docs/database.md
```diff
@@ -0,0 +1,21 @@
+# Interacting with Database
```
roborourke:
```suggestion
# Interacting with the Database
```

docs/database.md
```diff
@@ -0,0 +1,21 @@
...
+
+- You can log into MySQL and run any query in the default database (`wordpress`) by using:
```
roborourke:
```suggestion
You can log into MySQL and run any query in the default database `wordpress` by using:
```

docs/database.md
```diff
@@ -0,0 +1,21 @@
...
+
+```sh
+composer server db
+```
+
+- Use `composer server info` to retrieve MySQL info, connection details and forwarded ports:
```
roborourke:
```suggestion
Use `composer server db info` to retrieve the MySQL connection details including the hostname and port:
```

inc/composer/class-command.php
```diff
@@ -332,6 +338,63 @@ protected function shell( InputInterface $input, OutputInterface $output ) {
...
+
+				$db_info = <<<EOT
+<info>Root password</info>:  ${values['MYSQL_ROOT_PASSWORD']}
+
+<info>Database</info>:       ${values['MYSQL_PASSWORD']}
+<info>User</info>:           ${values['MYSQL_USER']}
+<info>Password</info>:       ${values['MYSQL_PASSWORD']}
+
+<comment>Version</comment>:        ${values['MYSQL_VERSION']}
+<comment>Ports</comment>:          ${ports_matches[1]}
```
roborourke: Can we split out the host and port here?

inc/composer/class-command.php
```diff
@@ -332,6 +338,63 @@ protected function shell( InputInterface $input, OutputInterface $output ) {
...
+EOT;
+				echo $output->write( $db_info ) . PHP_EOL;
```
roborourke: `echo` isn't needed here

roborourke: Actually why did you need to do the `PHP_EOL` outside of the call to `$output->write()`?

inc/composer/class-command.php
```diff
@@ -332,6 +338,63 @@ protected function shell( InputInterface $input, OutputInterface $output ) {
...
+					$values[ $env_variable[0] ] = $env_variable[1] ?? '';
+					return $values;
+				}, [] );
+
+				// Ports
+
+				// Find the container ID with Docker Compose
+				$db_container_id = shell_exec( "$base_command_prefix docker-compose ps -q db" );
+
+				// Retrrieve the forwarded ports using Docker and the container ID
```
roborourke:
```suggestion
				// Retrieve the forwarded ports using Docker and the container ID
```

inc/composer/class-command.php
igmoweb: It's an old code that I added during development. I just stripped it out

inc/composer/class-command.php
```diff
@@ -320,18 +335,108 @@ protected function logs( InputInterface $input, OutputInterface $output ) {
...
+
+				$db_info = <<<EOT
+<info>Root password</info>:  ${connection_data['MYSQL_ROOT_PASSWORD']}
+
+<info>Database</info>:       ${connection_data['MYSQL_DATABASE']}
+<info>User</info>:           ${connection_data['MYSQL_USER']}
+<info>Password</info>:       ${connection_data['MYSQL_PASSWORD']}
+
+<comment>Version</comment>:        ${connection_data['MYSQL_VERSION']}
+<comment>Ports</comment>:          ${connection_data['host']}:${connection_data['port']}
```
roborourke: I was thinking like this just to make it clearer, the `0.0.0.0` bit is the host name rather than a port
```suggestion
<comment>Host</comment>:           ${connection_data['host']}
<comment>Port</comment>:           ${connection_data['port']}
```

inc/composer/class-command.php
```diff
@@ -320,18 +335,135 @@ protected function logs( InputInterface $input, OutputInterface $output ) {
...
+
+EOT;
+
+				$output->write( $output_message );
+
+				$os = php_uname();
+				if ( strpos( $os, 'Darwin' ) !== false ) {
+					exec( "open $output_file_path", $null, $return_val );
+				} elseif ( strpos( $os, 'Windows' ) !== false ) {
+					exec( "start $output_file_path", $null, $return_val );
```
roborourke: Because Sequel Pro is mac only we don't need to do this check too, I think move the `strpos( $os, 'Darwin' ) !== false` check to the top of this subcommand section then output an error to say "This command is only supported on MacOS, use `composer server db info` to see the database connection details." and return early

inc/composer/class-command.php
```diff
@@ -320,18 +335,135 @@ protected function logs( InputInterface $input, OutputInterface $output ) {
...
+				foreach ( $connection_data as $field_name => $field_value ) {
+					$rpf_file_contents = preg_replace( "/(<%=\s)($field_name)(\s%>)/i", $field_value, $rpf_file_contents );
+				}
+				$output_file_path = getcwd() . '/sequel.spf';
+				file_put_contents( $output_file_path, $rpf_file_contents );
+
+				$output_message = <<<EOT
+The SPF file has been placed in <info>$output_file_path</info>.
+
+EOT;
```
roborourke: Given you fixed the file name suffix we can drop this output now

inc/composer/class-command.php
```diff
@@ -320,18 +335,135 @@ protected function logs( InputInterface $input, OutputInterface $output ) {
...
+				$output->write( $db_info );
+				break;
+			case 'spf':
+				$connection_data = $this->get_db_connection_data();
+				$rpf_file_contents = file_get_contents( getcwd() . '/vendor/altis/local-server/templates/sequel.xml' );
+				foreach ( $connection_data as $field_name => $field_value ) {
+					$rpf_file_contents = preg_replace( "/(<%=\s)($field_name)(\s%>)/i", $field_value, $rpf_file_contents );
+				}
+				$output_file_path = getcwd() . '/sequel.spf';
```
roborourke: This will get dumped into the root directory so would need to be git ignored, we can avoid this by writing to a temp file instead - the `/tmp` directory on linux based systems is cleared on every reboot so it's a good place to drop things that can be cleaned up automatically. Let's put this in `sprintf( '/tmp/%s.spf', $this->get_project_subdomain() )`

inc/composer/class-command.php
```diff
@@ -320,18 +335,135 @@ protected function logs( InputInterface $input, OutputInterface $output ) {
...
+				} elseif ( strpos( $os, 'Windows' ) !== false ) {
+					exec( "start $output_file_path", $null, $return_val );
+				}
+
+				if ( $return_val !== 0 ) {
+					$output->writeln( '<error>You must have Sequel Pro installed to use this command</error>' );
+				}
+
+				break;
+			default:
```
roborourke: Right now this would mean any additional string passed to `composer server db` would result in the db login behaviour which could be confusing in cases where perhaps someone makes a typo like `composer server db imfo` - it'll just log them in to the server.

Ideally we should check the `$db` argument is `null` like we're doing here in the `execute()` method:

https://github.com/humanmade/altis-local-server/pull/122/files#diff-a61d611804a0cdfa68973f824789bbeaR92-R98

The default behaviour when an unrecognised command is passed could then be to say "The subcommand 
{$db} is not recognised".

inc/composer/class-command.php
```diff
@@ -320,18 +335,135 @@ protected function logs( InputInterface $input, OutputInterface $output ) {
...
+<comment>Version</comment>:        ${connection_data['MYSQL_VERSION']}
+<comment>Host</comment>:          ${connection_data['HOST']}
+<comment>Port</comment>:          ${connection_data['PORT']}
+
+EOT;
+				$output->write( $db_info );
+				break;
+			case 'spf':
+				$connection_data = $this->get_db_connection_data();
+				$rpf_file_contents = file_get_contents( getcwd() . '/vendor/altis/local-server/templates/sequel.xml' );
```
roborourke: Any reason not to use `dirname( __DIR__, 2 )` or other relative path method? Whenever possible it's good to avoid writing out the full `vendor/namespace/package` paths because it's possible for projects to override the vendor directory location from the root `composer.json`

roborourke: Where is the "rpf" name coming from? Should that be "spf"?

igmoweb: There was no reason. I tried first a relative way but saw that `getcwd` was used often so I thought I was being consistent.

docs/database.md
```diff
@@ -0,0 +1,21 @@
...
+Use `composer server info` to retrieve MySQL info, connection details and forwarded ports:
+
+```sh
+Root password:  wordpress
+
+Database:       wordpress
+User:           wordpress
+Password:       wordpress
+
+Version:        5.7.26-1debian9
```
roborourke: Was this at the bottom? Think this just needs updating with the latest output.

inc/composer/class-command.php
```diff
@@ -320,18 +335,130 @@ protected function logs( InputInterface $input, OutputInterface $output ) {
...
+			"$command_prefix docker-compose exec -e COLUMNS=%d -e LINES=%d php /bin/bash",
 			$columns,
 			$lines
 		), $return_val );
 
 		return $return_val;
 	}
 
+	protected function db( InputInterface $input, OutputInterface $output ) {
+		$db = $input->getArgument( 'options' )[0] ?? '';
```
roborourke:
```suggestion
		$db = $input->getArgument( 'options' )[0] ?? null;
```
Just going to test this out, I think it's necessary for the default command to run.

inc/composer/class-command.php
```diff
@@ -320,18 +335,130 @@ protected function logs( InputInterface $input, OutputInterface $output ) {
...
+				$spf_file_contents = file_get_contents( dirname( __DIR__, 2 ) . '/templates/sequel.xml' );
+				foreach ( $connection_data as $field_name => $field_value ) {
+					$spf_file_contents = preg_replace( "/(<%=\s)($field_name)(\s%>)/i", $field_value, $spf_file_contents );
+				}
+				$output_file_path = sprintf( '/tmp/%s.spf', $this->get_project_subdomain() );
+				file_put_contents( $output_file_path, $spf_file_contents );
+
+				exec( "open $output_file_path", $null, $return_val );
+				if ( $return_val !== 0 ) {
+					$output->writeln( '<error>You must have Sequel Pro installed to use this command</error>' );
```
roborourke:
```suggestion
					$output->writeln( '<error>You must have <href=https://www.sequelpro.com/>Sequel Pro</> installed to use this command</error>' );
```

inc/composer/class-command.php
```diff
@@ -320,18 +335,130 @@ protected function logs( InputInterface $input, OutputInterface $output ) {
...
+<info>Password</info>:       ${connection_data['MYSQL_PASSWORD']}
+
+<comment>Version</comment>:        ${connection_data['MYSQL_VERSION']}
+<comment>Host</comment>:           ${connection_data['HOST']}
+<comment>Port</comment>:           ${connection_data['PORT']}
+
+EOT;
+				$output->write( $db_info );
+				break;
+			case 'spf':
```
roborourke:
```suggestion
			case 'spf':
			case 'sequel':
```
Spoke to Ryan and we'll use `sequel` as the primary name for this command

inc/composer/class-command.php
```diff
@@ -320,18 +335,130 @@ protected function logs( InputInterface $input, OutputInterface $output ) {
...
+				if ( $return_val !== 0 ) {
+					$output->writeln( '<error>You must have Sequel Pro installed to use this command</error>' );
+				}
+
+				break;
+			case null:
+				passthru( "$base_command mysql --database=wordpress --user=root -pwordpress", $return_val );
+				break;
+			default:
+				$output->writeln( "<error>The subcommand $db is not recognized</error>" );
```
roborourke:
```suggestion
				$output->writeln( "<error>The subcommand $db is not recognized</error>" );
				return 1;
```
Just adding a non zero exit code

inc/composer/class-command.php
```diff
@@ -42,6 +42,10 @@ protected function configure() {
 	exec -- <command>             eg: exec -- vendor/bin/phpcs
 Open a shell:
 	shell
+Database commands:
+	db                            Log into MySQL on the Database server
+	db spf                        Generates an SPF file for Sequel Pro
```
roborourke:
```suggestion
	db sequel                     Generates an SPF file for Sequel Pro
```

inc/composer/class-command.php
igmoweb: I've tested both ways and it always works but anyway, `null` looks slightly better

inc/composer/class-command.php
igmoweb: I've assigned `1` to `$return_val` instead. Although we already have one early return, it can get confusing if we write more code in that function.

inc/composer/class-command.php
igmoweb: I think that `<href` will only render properly in iTerm2 and GNOME. I don't use either of both and the output looks very ugly.  I have added the link between parenthesis if that's ok

inc/composer/class-command.php
roborourke: good shout, makes sense to me

inc/composer/class-command.php
roborourke: Ahh of course, the `switch` doesn't do strict matching. It's a `==` comparison 👍

inc/composer/class-command.php
roborourke: oh me either, yeah good call - I missed that bit in the docs

PR #125, opened 2020-02-14
Update Troubleshooting.md to add ES_MEM_LIMIT info

katmoody:
Include information about `ES_MEM_LIMIT` variable, how and why to update it, Per https://github.com/humanmade/altis-documentation/issues/113

Fixes https://github.com/humanmade/altis-documentation/issues/113

### DIFF
--- a/docs/elasticsearch.md
+++ b/docs/elasticsearch.md
@@ -49,10 +49,11 @@ Copy the mapped IP and port (highlighted in bold above) and use it to query Elas
 curl -XGET http://0.0.0.0:32871
 ```
 
-## Configuration updates when using ElasticSearch on Local Environment
+## ElasticSearch Memory Limit
 
-ElasticSearch uses more memory within the local server environment, which can also cause services to stop working,n particularly when using Ubuntu. Altis includes an environment variable which can change the default memory limit for ElasticSearch called `ES_MEM_LIMIT`. 
+ElasticSearch requires more memory on certain operating systems such as Ubuntu or when using Continuous Integration services. If ElasticSearch does not have enough memory it can cause other services to stop working. The Local Server supports an environment variable which can change the default memory limit for ElasticSearch called `ES_MEM_LIMIT`. 
 
-Using this variable with the `composer server start` command, users can update the memory limit for the current session only; it won't change the variable permenantly. An example might be: 
+You can set the `ES_MEM_LIMIT` variable in 2 ways:
 
-`ES_MEM_LIMIT=2g composer server start`
+- Set it globally eg: `export ES_MEM_LIMIT=2g`
+- Set it for the local server process only: `ES_MEM_LIMIT=2g composer server start`
--- a/docs/elasticsearch.md
+++ b/docs/elasticsearch.md
@@ -48,3 +48,17 @@ Copy the mapped IP and port (highlighted in bold above) and use it to query Elas
 ```
 curl -XGET http://0.0.0.0:32871
 ```
+
+## Configuration updates when using ElasticSearch on Local Environment
+
+ElasticSearch uses more memory within the local server environment, which can also cause services to stop working,n particularly when using Ubuntu. Altis includes an environment variable which can change the default memory limit for ElasticSearch called `ES_MEM_LIMIT`. 
+
+### Change variable during current session only
+
+Using this variable with the `composer server start` command, users can update the memory limit for the current session only; it won't change the variable permenantly. An example might be: 
+
+`ES_MEM_LIMIT=2g composer server start`
+
+### Change variable permanently
+
+Users can also define that environment variable on the container's config file if they wanted a permanent fix. For instance, editing or adding the variable to the "environment" section of the [/docker-compose.yml](https://github.com/humanmade/altis-local-server/blob/master/docker/docker-compose.yml) Altis Local Server setup.
--- a/docs/elasticsearch.md
+++ b/docs/elasticsearch.md
@@ -58,7 +58,3 @@ ElasticSearch uses more memory within the local server environment, which can al
 Using this variable with the `composer server start` command, users can update the memory limit for the current session only; it won't change the variable permenantly. An example might be: 
 
 `ES_MEM_LIMIT=2g composer server start`
-
-### Change variable permanently
-
-Users can also define that environment variable on the container's config file if they wanted a permanent fix. For instance, editing or adding the variable to the "environment" section of the [/docker-compose.yml](https://github.com/humanmade/altis-local-server/blob/master/docker/docker-compose.yml) Altis Local Server setup.
--- a/docs/elasticsearch.md
+++ b/docs/elasticsearch.md
@@ -53,8 +53,6 @@ curl -XGET http://0.0.0.0:32871
 
 ElasticSearch uses more memory within the local server environment, which can also cause services to stop working,n particularly when using Ubuntu. Altis includes an environment variable which can change the default memory limit for ElasticSearch called `ES_MEM_LIMIT`. 
 
-### Change variable during current session only
-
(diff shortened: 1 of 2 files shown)


docs/elasticsearch.md
```diff
@@ -48,3 +48,17 @@ Copy the mapped IP and port (highlighted in bold above) and use it to query Elas
...
+
+### Change variable during current session only
+
+Using this variable with the `composer server start` command, users can update the memory limit for the current session only; it won't change the variable permenantly. An example might be: 
+
+`ES_MEM_LIMIT=2g composer server start`
+
+### Change variable permanently
+
+Users can also define that environment variable on the container's config file if they wanted a permanent fix. For instance, editing or adding the variable to the "environment" section of the [/docker-compose.yml](https://github.com/humanmade/altis-local-server/blob/master/docker/docker-compose.yml) Altis Local Server setup.
```
roborourke: We definitely don't want to recommend editing any core files so you can remove this bit :)

docs/elasticsearch.md
```diff
@@ -48,3 +48,17 @@ Copy the mapped IP and port (highlighted in bold above) and use it to query Elas
 ```
 curl -XGET http://0.0.0.0:32871
 ```
+
+## Configuration updates when using ElasticSearch on Local Environment
```
roborourke: Sorry I must've made an error on the issue, the issue isn't to do with the local environment but rather continuous integration environments like Travis and CircleCI and Ubuntu operating systems.

The purpose of this section is just to document the environment variable so can you make the title:
```suggestion
## ElasticSearch Memory Limit
```

docs/elasticsearch.md
```diff
@@ -48,3 +48,17 @@ Copy the mapped IP and port (highlighted in bold above) and use it to query Elas
...
+
+ElasticSearch uses more memory within the local server environment, which can also cause services to stop working,n particularly when using Ubuntu. Altis includes an environment variable which can change the default memory limit for ElasticSearch called `ES_MEM_LIMIT`.
```
roborourke:
```suggestion
ElasticSearch requires more memory on certain operating systems such as Ubuntu or in Continuous Integration services. When ElasticSearch does not have enough memory it can cause other services to stop working. The Local Server supports an environment variable which can change the default memory limit for ElasticSearch called `ES_MEM_LIMIT`. 
```

docs/elasticsearch.md
```diff
@@ -48,3 +48,17 @@ Copy the mapped IP and port (highlighted in bold above) and use it to query Elas
...
 curl -XGET http://0.0.0.0:32871
 ```
+
+## Configuration updates when using ElasticSearch on Local Environment
+
+ElasticSearch uses more memory within the local server environment, which can also cause services to stop working,n particularly when using Ubuntu. Altis includes an environment variable which can change the default memory limit for ElasticSearch called `ES_MEM_LIMIT`. 
+
+### Change variable during current session only
+
+Using this variable with the `composer server start` command, users can update the memory limit for the current session only; it won't change the variable permenantly. An example might be:
```
roborourke: We're addressing the individual here, they're reading the docs to find out how to do something so it's a bit odd to read "users can update". This is something the reader needs to do.

docs/elasticsearch.md
```diff
@@ -48,3 +48,17 @@ Copy the mapped IP and port (highlighted in bold above) and use it to query Elas
 ```
 curl -XGET http://0.0.0.0:32871
 ```
+
+## Configuration updates when using ElasticSearch on Local Environment
+
+ElasticSearch uses more memory within the local server environment, which can also cause services to stop working,n particularly when using Ubuntu. Altis includes an environment variable which can change the default memory limit for ElasticSearch called `ES_MEM_LIMIT`. 
+
+### Change variable during current session only
```
roborourke: I think we can drop this heading, it's a short section and without the last section a bit unnecessary I think.

docs/elasticsearch.md
roborourke:
```suggestion
You can set the `ES_MEM_LIMIT` variable in 2 ways:

- Set it globally eg: `export ES_MEM_LIMIT=2g`
- Set it for the local server process only: `ES_MEM_LIMIT=2g composer server start`
```

docs/troubleshooting.md
```diff
@@ -28,6 +28,20 @@ In the docker GUI go to the "Preferences" pane, then the "Advanced" tab and move
 
 ![Docker Advanced Settings](./assets/docker-gui-advanced.png)
 
+## Configuration updates when using ElasticSearch on Local Environment
```
roborourke: Because this is troubleshooting try to think about the questions someone might be asking themselves or searching for when writing headings eg. "Services failing to start or randomly stopping"

docs/elasticsearch.md
katmoody: This was my error - I had in my notes it was for the continuous integration environments but when I rewrote I used a section just below that, basically cutting it from the notes. Sorry about that.

docs/elasticsearch.md
katmoody: instead of "in Continuous Integration services" I think it should be "when using ..." as it's not actually "in" those services but using or utilizing them.

docs/elasticsearch.md
katmoody: So maybe instead
```Update the memory limit for the current session only by using this variable with the `composer server start` command; it won't change the variable permanently. For example: ```

File: docs/elasticsearch.md (hunk 3 above)
Comment (roborourke) [reply docs/elasticsearch.md]: @katmoody yep - I think I used that term as we'd be running elasticsearch within the server that the CI service gives us but you're right. You can edit my suggestion here then commit it

File: docs/elasticsearch.md (hunk 2 above)
Comment (roborourke) [reply docs/elasticsearch.md]: With the suggestions you can commit them from the UI directly here - looks like you marked it resolved before doing that :)

File: docs/elasticsearch.md [hunk 7]
```diff
@@ -48,3 +48,11 @@ Copy the mapped IP and port (highlighted in bold above) and use it to query Elas
...
 curl -XGET http://0.0.0.0:32871
 ```
+
+## Configuration updates when using ElasticSearch on Local Environment
+
+ElasticSearch uses more memory within the local server environment, which can also cause services to stop working,n particularly when using Ubuntu. Altis includes an environment variable which can change the default memory limit for ElasticSearch called `ES_MEM_LIMIT`. 
+
+Using this variable with the `composer server start` command, users can update the memory limit for the current session only; it won't change the variable permenantly. An example might be: 
+
+`ES_MEM_LIMIT=2g composer server start`
```
roborourke:
```suggestion
```

docs/troubleshooting.md
```diff
@@ -28,6 +28,20 @@ In the docker GUI go to the "Preferences" pane, then the "Advanced" tab and move
...
+
+### Change variable during current session only
+
+Using this variable with the `composer server start` command, users can update the memory limit for the current session only; it won't change the variable permenantly. An example might be: 
+
+`ES_MEM_LIMIT=2g composer server start`
+
+### Change variable permanently
+
+Users can also define that environment variable on the container's config file if they wanted a permanent fix. For instance, editing or adding the variable to the "environment" section of the [/docker-compose.yml](https://github.com/humanmade/altis-local-server/blob/master/docker/docker-compose.yml) Altis Local Server setup.
```
roborourke:
```suggestion
ElasticSearch requires more memory on certain operating systems such as Ubuntu or when using Continuous Integration services. If ElasticSearch does not have enough memory it can cause other services to stop working. The Local Server supports an environment variable which can change the default memory limit for ElasticSearch called `ES_MEM_LIMIT`. 

You can set the `ES_MEM_LIMIT` variable in 2 ways:

- Set it globally eg: `export ES_MEM_LIMIT=2g`
- Set it for the local server process only: `ES_MEM_LIMIT=2g composer server start`
```

PR #143, opened 2020-03-20
WIP: Add support for external altis.dev domains

roborourke:
Because *.altis.dev maps to localhost the PHP container cant make external requests via curl to the other services. This sets their IP addresses to the docker host IP.

This is one hacky approach, but I'm confident we can find a better one. Just putting this up for visibility.


roborourke:
proper solution in #144

PR #144, opened 2020-03-20
Network PHP container to allow access to service hostnames

joehoyle:
Now PHP uses sockets instead of port 9000 for the Nginx -> PHP connection, we can put PHP on the main proxy network. This means it's then easy to link in the hostnames of our services to the traefik HTTP server.

This means we can make requests from PHP to https://project.altis.dev/ as well as https://pinpoint-project.altis.dev/ etc.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -19,10 +19,15 @@ x-php: &php
     - "s3:s3.localhost"
   external_links:
     - "proxy:${COMPOSE_PROJECT_NAME:-default}.altis.dev"
+    - "proxy:pinpoint-${COMPOSE_PROJECT_NAME:-default}.altis.dev"
+    - "proxy:cognito-${COMPOSE_PROJECT_NAME:-default}.altis.dev"
   volumes:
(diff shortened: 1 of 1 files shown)


docker/docker-compose.yml
```diff
@@ -19,10 +19,15 @@ x-php: &php
     - "s3:s3.localhost"
   external_links:
     - "proxy:${COMPOSE_PROJECT_NAME:-default}.altis.dev"
+    - "proxy:pinpoint-${COMPOSE_PROJECT_NAME:-default}.altis.dev"
+    - "proxy:cognito-${COMPOSE_PROJECT_NAME:-default}.altis.dev"
```
roborourke: If we're adding these we should explicitly add `elasticsearch` and `s3` too
```suggestion
    - "proxy:cognito-${COMPOSE_PROJECT_NAME:-default}.altis.dev"
    - "proxy:elasticsearch-${COMPOSE_PROJECT_NAME:-default}.altis.dev"
    - "proxy:s3-${COMPOSE_PROJECT_NAME:-default}.altis.dev"
```

Issue #142, opened 2020-03-17
Document the process of importing uploads to FakeS3 for use in Tachyon


goldenapples:
I [asked on Slack earlier](https://hmn.slack.com/archives/C3CKS8LLR/p1584464726497000) about help troubleshooting Tachyon images in my local-server install, where a uploads directory I had copied into my `/content` directory from a production export were not available in Tachyon.

The part of the process which I didn't understand beforehand was that the `content/uploads` directory isn't automatically synced to the fake s3 service container which tachyon accesses. I was able to "upload" my uploads directory by running `s3-uploads upload-directory`:

```
composer server cli -- s3-uploads upload-directory content/uploads
```

This would be good to explain in the documentation. I'll come back to do this soon, just leaving a note here for myself.

--------------

Acceptance criteria:

- [ ] Section added to troubleshooting with the title "Files in uploads directory not available via Tachyon or direct URL" with an explanation of why and the command to run

roborourke:
I think we're going to want a full guide for importing existing data for both local envs. Will add separate issues in the docs repo for that.

roborourke:
Since moving to Minio we've added commands to help with upload imports and documented those so we can close this one out now.

PR #145, opened 2020-03-21
Use Minio for S3 storage

### LINKED ISSUE #140: Add ability to map uploads folder
@roborourke Let me see if I can explain this properly for what I _think_ is needed with local-server.

So the other day I needed to access the uploads folder inside of s3 docker, however because this isn't mapped to any folder on the host machine. I had to go and change the docker yml file locally, to map it to the uploads folder on the host machine.

My recommendation here is that do the same thing by default: Map the S3 docker instance to the uploads folder, so we have access to the contents inside the uploads folder. This gives the developer the ability to add assets when using assets from production etc.

Hopefully this makes sense, happy to discuss further.

roborourke:
This update switches from fakes3 to minio plus the minio client (mc). This
   allows us to properly support virtualhost style routing as per AWS where
   the subdomain selects the bucket.

   By using the `mc` docker image we can remove some not too robust code that
   was designed to ensure the bucket was created by uploading a file and then
   removing it.

   Additionally we can use `mc` to run a synchronisation service. Currently I
   have this running to sync from s3 to `content/uploads`. During startup eg.
   `composer serve` any existing files in `content/uploads` are synced to s3.
   Unfortunately getting this to work well in a bidirectional way is almost
   impossible. So we may want to think about our preferred direction to do
   synchronisation by default and perhaps even add a command to run the sync
   from the host to s3.

Fixes #140

Related to #142

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -16,7 +16,7 @@ x-php: &php
   image: humanmade/altis-local-server-php:3.1.0
   links:
     - "db:db-read-replica"
-    - "s3:s3.localhost"
+    - "minio:s3.localhost"
   external_links:
     - "proxy:${COMPOSE_PROJECT_NAME:-default}.altis.dev"
     - "proxy:pinpoint-${COMPOSE_PROJECT_NAME:-default}.altis.dev"
@@ -47,7 +47,7 @@ x-php: &php
     ELASTICSEARCH_HOST: elasticsearch
     ELASTICSEARCH_PORT: 9200
     AWS_XRAY_DAEMON_HOST: xray
-    S3_UPLOADS_ENDPOINT: http://s3.localhost:8000/
+    S3_UPLOADS_ENDPOINT: http://s3.localhost:9000/
     S3_UPLOADS_BUCKET: s3-${COMPOSE_PROJECT_NAME:-default}
     S3_UPLOADS_BUCKET_URL: https://s3-${COMPOSE_PROJECT_NAME:-default}.altis.dev
     S3_UPLOADS_KEY: not-needed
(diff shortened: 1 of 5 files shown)


docker/docker-compose.yml
```diff
@@ -153,19 +162,50 @@ services:
...
+    <<: *mc
+    depends_on:
+      - s3-make-bucket
+    command: ["policy", "set", "public", "local/s3-${COMPOSE_PROJECT_NAME:-default}"]
+  s3-sync-up:
+    <<: *mc
+    depends_on:
+      - s3-make-bucket
+    command: ["mirror", "--watch", "--overwrite", "local/s3-${COMPOSE_PROJECT_NAME:-default}", "/content"]
+  s3-sync-down:
```
joehoyle: Do these all need to be top level services? It doesn't seem like the docker-compose file should be used for this maybe?

docs/README.md
```diff
@@ -45,7 +45,9 @@ The subdomain used for the project can be configured via the `modules.local-serv
 
 ## Available Commands
 
-* `composer server start [--xdebug]` - Starts the containers. If the `--xdebug` option is passed the PHP container will have XDebug enabled. To switch off XDebug run this command again without the `--xdebug` option.
+* `composer server start [--xdebug]` - Starts the containers.
+  * If the `--xdebug` option is passed the PHP container will have XDebug enabled. To switch off XDebug run this command again without the `--xdebug` option.
+  * Any files in `content/upload` are synced to the S3 server on start up.
```
joehoyle: I don't think we should add this behaviour for `start` by default, at least now. I'd say we should keep it as it's own command.

docs/README.md
```diff
@@ -57,3 +59,4 @@ The subdomain used for the project can be configured via the `modules.local-serv
 * `composer server db` - Logs into MySQL on the DB container.
   * `composer server db info` - Print MySQL connection details.
   * `composer server db sequel` - Opens a connection to the database in [Sequel Pro](https://sequelpro.com).
+* `composer server sync` - Sync files from `content/uploads` to the s3 container.
```
joehoyle: I think we need a more clearer name than `sync`, needs to speak to uploads. `import-uploads` maybe? It doesn't appear to be a sync operation, because it's only one way, with overriting enabled?

docker/docker-compose.yml
roborourke: I did wonder the same but it works really well at least. The other option to have a wrapper function that does `docker run minio/mc ...` and call that from php however for the long running mirror command with `—watch` it’s essentially a service.

docs/README.md
roborourke: Why so in your view? I don’t disagree necessarily, my thinking was that it’s just one less thing to think about when starting the server.

To be fair it is weird to have docker compose services that immediately exit.

docs/README.md
roborourke: It’s essentially `rsync`, so not sure sync necessarily implies 2 way either but agree on the naming. Import implies the specific direction this is running in so I think that makes sense.

roborourke: Updated the command name

docs/README.md
roborourke: I removed this, I think you're right, it's less magical.

docker/docker-compose.yml
roborourke: Found a way to get the best of all worlds. The main benefit to the previous appraoch was using docker-compose `depends_on` to ensure the bucket was created and the policy set however it's possible to use `entrypoint` instead and execute a string of commands. Removes the need for any messing around in PHP still.

PR #146, opened 2020-03-27
Change exec function to run cli commands as www-data instead of nobody

nathanielks:
If a user goes to run a command that would modify the filesystem (such as manually installing a plugin), they will be denied access as `nobody` does not have sufficient permissions to write to /usr/src/app. Changing this to www-data will grant those permissions.


nathanielks:
I believe v3 is being treated than v2 as I see v3 has an `exec()` function but v2 still uses a `cli()` function. I'll make a separate PR to update v2, though let me know if that's the right course of action.

joehoyle:
YUp that's the right approach, this just needs the backport-v3 label adding to get it into v3.

PR #147, opened 2020-03-27
V2: Change exec function to run cli commands as www-data instead of nobody

nathanielks:
If a user goes to run a command that would modify the filesystem (such as manually installing a plugin), they will be denied access as nobody does not have sufficient permissions to write to /usr/src/app. Changing this to www-data will grant those permissions.


nathanielks:
I'll do a separate PR for v1.

PR #154, opened 2020-05-08
Update PHP Container 3.2.0

roborourke:
This brings WP CLI up to date at 2.4.0.


roborourke:
Merging #155 into this should fix the linter problems

PR #155, opened 2020-05-08
Ignore PHPCS errors in composer command

roborourke:
Calls to functions like `file_get_contents()` etc... are being flagged as errors however in the composer command we are outside the scope of WordPress so there's no problem using them.


rmccue:
I did remove them in the HM-Required ruleset, and didn't add to HM, so they'll go away soon.

PR #156, opened 2020-05-14
Only set the X-Ray daemon address if it isn't already defined

joehoyle:
This avoids a notice in some cases.


roborourke:
Does this need backporting?

joehoyle:
Nah i think it's fine, just a small enhacement

PR #159, opened 2020-05-14
Generate docker-compose.yml file in PHP

joehoyle:
We are reaching the limitations of what we can do with a static docker-compose file. There's no way to optionally disable services for example, or provide configuration for the containers from the `composer.json` or CLI flags. This moves the definition of the docker-compose file into PHP, and will dump it out to a docker-compose.yml file for use in `docker-compose`.


roborourke:
We discussed in slack briefly but adding my high level thoughts here for the records:

I'm +1 on this approach, gives a huge amount more flexibility to tailor the environment through config and should simplify passing environment variables around as there's only one level we need to read them at this way.

nathanielks:
This sounds great 👍

joehoyle:
@roborourke Ok I updated this with the latest changes in master, and I think we should go with it before it gets stale again. I've ro-organized the generated file, and also committed an example build over the old docker-composer.yml, so we can look at the diff here to see what it's going to generate differently. A second pair of eyes on the generated docker-compose.yml would be great!

joehoyle:
With the latest change, you can also put ```        "local-server": {
          "elasticsearch": false,
          "cavalcade": false,
          "analytics": false
        },```

In your project's config, to have a much slimmer local env. We _could_ have these instead inherit from the main modules that use them, so if you set `modules.analytics.enabled = false` we could default also to not running those. I feel like this might be a better way to go (or support it as the default, and allow the `modules.local-server.analytics` as a higher priority override.

nathanielks:
eg load the config that exists in the project and modify just the fields that need modification?

joehoyle:
> eg load the config that exists in the project and modify just the fields that need modification?

I don't think long-term this will be more maintainable. I think `php => yaml` is going to be better than `yaml => php => yaml`. Also, there are subtleties in output formatting that won't make it _exactly_ the same as the original docker-compose.yml, so I don't think it helps a lot in being able to cleanly diff/discern exactly what was changed.

roborourke:
> With the latest change, you can also put
> 
> ```
>           "elasticsearch": false,
>           "cavalcade": false,
>           "analytics": false
> },
> ```
> 
> In your project's config, to have a much slimmer local env. We _could_ have these instead inherit from the main modules that use them, so if you set `modules.analytics.enabled = false` we could default also to not running those. I feel like this might be a better way to go (or support it as the default, and allow the `modules.local-server.analytics` as a higher priority override.

That does make sense but it'll be somewhat complicated I think as we'd need to get the merged config to ensure the local env overrides are applied too. I'm not sure we can do that easily in composer commands as we've had to just read directly from `composer.json` to do these in the past but I might be wrong if the main Altis core namespace is autoloaded.

docker/docker-compose.yml
```diff
@@ -1,285 +1,333 @@
...
       AWS_SECRET_ACCESS_KEY: YOUR_SECRET_HERE
       AWS_REGION: us-east-1
   cavalcade:
-    <<: *php
     entrypoint:
       - /usr/local/bin/cavalcade
-    user: "nobody:nobody"
-    restart: unless-stopped
+    user: 'nobody:nobody'
+    restart: on-failure
```
roborourke, line 118: This value seems to have changed compared to the previous

docker/docker-compose.yml
```diff
@@ -1,285 +1,333 @@
...
+      - '/Users/joe/altis/dev/packages/local-server/docker/minio.json:/root/.mc/config.json'
+      - '/Users/joe/altis/dev/content/uploads:/content/uploads:delegated'
     links:
       - s3
-    entrypoint: >
-      /bin/sh -c "
-      mc mb -p local/s3-${COMPOSE_PROJECT_NAME:-default}
-      && mc policy set public local/s3-${COMPOSE_PROJECT_NAME:-default}
-      && mc mirror --watch --overwrite local/s3-${COMPOSE_PROJECT_NAME:-default} /content"
+    entrypoint: '/bin/sh -c " mc mb -p local/s3-dev && mc policy set public local/s3-dev && mc mirror --watch local/s3-dev /content"'
```
roborourke, line 245: Missing the `--overwrite` flag here

docker/docker-compose.yml
```diff
@@ -1,285 +1,333 @@
...
       MH_UI_WEB_PATH: mailhog
   cognito:
-    <<: *analytics
-    image: humanmade/local-cognito:1.0.0
+    ports:
+      - '3000'
+    networks:
+      - proxy
+      - default
+    restart: on-failure
```
roborourke, line 284: Need to set this to `unless-stopped`

inc/composer/class-docker-compose-generator.php
```diff
@@ -0,0 +1,513 @@
...
+	}
+	protected function get_service_cavalcade() : array {
+		return [
+			'cavalcade' => array_merge(
+				[
+					'entrypoint' => [
+						'/usr/local/bin/cavalcade',
+					],
+					'user' => 'nobody:nobody',
+					'restart' => 'on-failure',
```
roborourke, line 117:
```suggestion
					'restart' => 'unless-stopped',
```

inc/composer/class-docker-compose-generator.php
```diff
@@ -0,0 +1,513 @@
...
+					'8080',
+				],
+				'labels' => [
+					'traefik.frontend.priority=1',
+					'traefik.port=8080',
+					'traefik.protocol=https',
+					'traefik.docker.network=proxy',
+					"traefik.frontend.rule=HostRegexp:{$this->hostname},{subdomain:[a-z.-_]+}.{$this->hostname}",
+				],
+			]
```
roborourke, line 148:
```suggestion
			],
```

inc/composer/class-docker-compose-generator.php
```diff
@@ -0,0 +1,513 @@
...
+					's3',
+				],
+				'volumes' => [
+					"{$this->config_dir}/minio.json:/root/.mc/config.json",
+					"{$this->root_dir}/content/uploads:/content/uploads:delegated",
+				],
+				'links' => [
+					's3',
+				],
+				'entrypoint' => "/bin/sh -c \" mc mb -p local/s3-{$this->project_name} && mc policy set public local/s3-{$this->project_name} && mc mirror --watch local/s3-{$this->project_name} /content\"",
```
roborourke, line 324:
```suggestion
				'entrypoint' => "/bin/sh -c \"mc mb -p local/s3-{$this->project_name} && mc policy set public local/s3-{$this->project_name} && mc mirror --watch --overwrite local/s3-{$this->project_name} /content\"",
```

inc/composer/class-docker-compose-generator.php
```diff
@@ -0,0 +1,513 @@
...
+		return [
+			'cognito' => [
+				'ports' => [
+					'3000',
+				],
+				'networks' => [
+					'proxy',
+					'default',
+				],
+				'restart' => 'on-failure',
```
roborourke, line 389:
```suggestion
				'restart' => 'unless-stopped',
```

inc/composer/class-docker-compose-generator.php
```diff
@@ -0,0 +1,513 @@
...
+			],
+			'pinpoint' => [
+				'ports' => [
+					'3000',
+				],
+				'networks' => [
+					'proxy',
+					'default',
+				],
+				'restart' => 'on-failure',
```
roborourke, line 406:
```suggestion
				'restart' => 'unless-stopped',
```

inc/composer/class-docker-compose-generator.php
```diff
@@ -0,0 +1,513 @@
...
+
+		if ( $this->get_config()['elasticsearch'] ) {
+			$services = array_merge( $services, $this->get_service_elasticsearch() );
+		}
+
+		$services = array_merge(
+			$services,
+			$this->get_service_s3(),
+			$this->get_service_tachyon(),
+			$this->get_service_mailhog(),
```
roborourke, line 459:
```suggestion
			$this->get_service_mailhog()
```

inc/composer/class-docker-compose-generator.php
```diff
@@ -0,0 +1,513 @@
...
+		);
+		if ( $this->get_config()['analytics'] ) {
```
roborourke, line 461: Analytics depends on elasticsearch so that should be factored in here
```suggestion
		if ( $this->get_config()['analytics'] && $this->get_config()['elasticsearch'] ) {
```

inc/composer/class-docker-compose-generator.php
```diff
@@ -0,0 +1,513 @@
...
+			$services = array_merge( $services, $this->get_service_analytics() );
+		}
+		if ( $this->get_config()['elasticsearch'] ) {
```
roborourke, line 464: Better to allow kibana to be required or not, eg. we don't need it in Travis or CI envs, at least not by default.
```suggestion
		if ( $this->get_config()['kibana'] && $this->get_config()['elasticsearch'] ) {
```

inc/composer/class-docker-compose-generator.php
```diff
@@ -0,0 +1,513 @@
...
+	protected function get_config( $module = 'local-server' ) : array {
+		// @codingStandardsIgnoreLine
+		$json = file_get_contents( $this->root_dir . DIRECTORY_SEPARATOR . 'composer.json' );
+		$composer_json = json_decode( $json, true );
+
+		$config = ( $composer_json['extra']['altis']['modules'][ $module ] ?? [] );
+		$defaults = [
+			'analytics' => true,
+			'cavalcade' => true,
+			'elasticsearch' => true,
```
roborourke, line 506:
```suggestion
			'elasticsearch' => true,
			'kibana' => true,
```

roborourke (changes requested):
I've added in the missing bits / out of date bits as suggestions but this looks good on the whole. Couple of other things:

- Is `db-data:` equivalent to `db-data: null` eg. in the volumes section?
- The docker compose YAML should be output during the `start` command. Right now you have to dump the YAML then start the server
   - I've generally used the vendor directory as a place to dump generated files like this as it's guaranteed to be git ignored
- The class file code style needs to reflect the rest of what we do, eg. doc blocks and some more inline comments might be useful where there are `if` statements used in the method that deals with config

nathanielks:
> so I don't think it helps a lot in being able to cleanly diff/discern exactly what was changed.

I recognize this is a matter of preference, but I do think using a template file would simplify development and provide a far cleaner diff (I don't think we'd even need a yaml parser, necessarily). In essence, I would advocate for what we're doing in our Terraform configuration. We have template files that posses placeholder tokens that get replaced by the [`template_file`](https://registry.terraform.io/providers/hashicorp/template/latest/docs/data-sources/file) data source. I see a lot of what is effectively boilerplate in the current array syntax, and for someone whose hands are injured, I try to reduce boilerplate where I can. Using a template file would reduce the effort to a `file_get_contents()`, `sprintf()`, and `file_put_contents()` 🙂

nathanielks:
That said, I definitely won't block this PR but I wanted to hopefully make our lives easier!

joehoyle:
>  Using a template file would reduce the effort to file_get_contents(), sprintf(), and file_put_contents() 🙂

The intention of this PR is to open up a lot more flexibility than what can be done this way. The conditional loading of services (which also can change the healthecks) is one example, but I believe taking this approach will open up more possibilities too.

nathanielks:
By all means, carry on!

joehoyle:
@roborourke should we get this into b2, or go with v6 at this point?

roborourke:
I think v6 as it's a fundamental change and we should keep to the feature / code freeze to keep things manageable. We would also want to add things that integrate and make use of it at the same time ideally such as removing unneeded services in CI.

roborourke (approved):

PR #160, opened 2020-05-20
Add a section on setting vm.max_map_count

roborourke:
On Ubuntu and other Linux systems the ElasticSearch container is in production mode which requires the Docker VM setting `vm.max_map_count` to be a minimum of `262144`.

https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html#_set_vm_max_map_count_to_at_least_262144

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -27,6 +27,8 @@ To access your site visit: https://my-site.altis.dev/
 
 Visiting your site's URL should now work. Visit `/wp-admin/` and login with `admin` / `admin` to get started!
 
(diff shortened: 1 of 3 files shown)


roborourke:
Props to @kadamwhite for running into this and persevering with debugging via slack.

joehoyle:
Hmm how come we are running ES in prod mode in those cases? Can we change that?

roborourke:
@joehoyle it’s nothing we’re doing, there’s something to do with the way docker engine runs when it’s on Linux - I’m not sure we can control it. I’ll look into it though as I agree - this is also going to affect the Travis environment

PR #169, opened 2020-06-11
Fix memory usage in ES

### LINKED ISSUE #115: Elasticsearch memory limit
Currently we give the elasticsearch service 1gb of memory for the docker container. This often results in start failures, where increasing it to 2gb will usually resolve it.

Per the docs on https://elk-docker.readthedocs.io/

> A minimum of 4GB RAM assigned to Docker
> Elasticsearch alone needs at least 2GB of RAM to run.

Apart from the fact that I think it's crazy you need 2GB of RAM to run ES, we should either increase our mem size to 2GB for ES, or look into ways to run ES with less memory.

rmccue:
Fixes #115.

#44 added a new environment option, which overrode the existing one where we set the memory limit for Java. This meant that Java was trying to use more memory than the container was actually allocated, getting it killed by the OOM killer.

Additionally, because the [config sets network.host](https://github.com/blacktop/docker-elasticsearch-alpine/blob/765493a6597d20b583e8b82536d0ec5fcc357e32/6.3/config/elastic/elasticsearch.yml#L1), ES [switches into production mode](https://www.elastic.co/guide/en/elasticsearch/reference/master/bootstrap-checks.html#dev-vs-prod-mode). Notably, this turns the bootstrap checks from warnings into exceptions, which means that ES doesn't launch due to the low `vm.max_map_count` value. To switch it back into development mode, we need to change the discovery type back to single node (which is another env var we can set here).

All up, this should fix the ES memory problems and other related startup problems; it should also now not be necessary to increase Docker's memory limit.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -144,8 +144,6 @@ services:
       interval: 5s
       timeout: 5s
       retries: 10
-    environment:
-      ES_JAVA_OPTS: "-Xms512m -Xmx512m"
     labels:
       - "traefik.port=9200"
       - "traefik.protocol=http"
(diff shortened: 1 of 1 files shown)


roborourke:
Now available in version 3.0.4

PR #171, opened 2020-06-22
Stop containers restarting if docker daemon rebooted

### LINKED ISSUE #164: Remove restart on failure directive on Cavalcade container
There's a problem that was previously worked around by setting the Cavalcade container to always restart.

This is because it would exit if WP wasn't installed yet which isn't possible because the containers must all be started before WP can be installed.

See #81 for background.

The changes in Cavalcade v2 should help to mitigate this issue so we need to test that and if successful remove the restart on failure configuration.

@kadamwhite does this more or less cover the issue you noticed with the Cavalcade container not stopping? I thought restart on failure should allow it to be stopped successfully.

roborourke:
The `on-failure` restart policy meant that if you shut down or rebooted your laptop without first halting the local server then the cavalcade and analytics services would immediately try to start back up again without the rest of the required services to function.

This is primarily an issue for Cavalcade which runs every second, fails and restarts every second.

Changing to `unless-stopped` will ensure these services stay up for the durection of the local server instance or until the docker daemon is restarted such as after a shutdown / reboot.

Fixes #164

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -67,7 +67,7 @@ x-analytics: &analytics
   networks:
     - proxy
     - default
-  restart: on-failure
+  restart: unless-stopped
 
 services:
   db:
@@ -124,7 +124,7 @@ services:
     entrypoint:
       - /usr/local/bin/cavalcade
     user: "nobody:nobody"
-    restart: on-failure
+    restart: unless-stopped
   elasticsearch:
     image: humanmade/altis-local-server-elasticsearch:3.0.0
     ulimits:


kadamwhite:
Testing results (on Ubuntu 20.04):

- I ran `composer local-server stop` to shut down Altis,
- Ran `composer update` to pull latest across the board,
- Replaced the `altis/local-server` directory with a checkout of this branch,
- Ran `composer local-server start`, and
- Rebooted (without shutting anything down).

Upon reboot, several containers were initially active.

- I turned them all off with `docker container stop :id1 :id2 :id3...`, then 
- Rebooted again

`docker container ls` now _still_ shows Cavalcade, after two reboots.

My methodology may be flawed but this does not at first blush seem to be solving the issue I described in Slack.

kadamwhite:
Update:


Tried,

- composer local-server stop
- docker container prune
- composer local-container start
- composer local-server stop
- (Reboot)

Mercifully, after that it finally stopped. 

And if I do reboot with it running, several containers come back on reboot but if I manually stop them they seem to stay stopped. So, seems fixed!

Issue #123, opened 2020-02-10
Turn display_errors on by default


joehoyle:
Currently we use the `php.ini` from production, which has `display_errors = Off`. This means we don't get errors showing by default for fatals, which makes things like `wp cli` commands failing super difficult to debug.

I thinl we should change this to default on. I think we need to do this in the _actual_ php.ini because setting it in code falls into the "error happens before this code" problem.

Acceptance criteria:

- [x] Remove the line https://github.com/humanmade/altis-local-server/blob/461953201911a88f092969da24f003f89d25b081/inc/namespace.php#L52
- [x] In the module's `wp-config.php` file check if it's the CLI environment and if so call `ini_set( 
'display_errors', 'on' );`

To test restart the local server to apply the changes and confirm this works by calling an undefined function at the top of the `wp-config.php` file and running a CLI command. You can also `var_dump( ini_get( 'display_errors' ) )` to confirm the new default value.

roborourke:
@joehoyle this can go in `docker/php.ini` - the ini file is mounted into the PHP docker image and automatically included. Currently used to enable mailhog.

joehoyle:
Oh wow nice

rmccue:
This should be set for the CLI environment (i.e. php-cli), but not for the web, as the logs command should be used for that instead.

FWIW, it's already turned on in the local-server bootstrap file, but we should actually remove this so that errors aren't displayed in a HTML context:
https://github.com/humanmade/altis-local-server/blob/461953201911a88f092969da24f003f89d25b081/inc/namespace.php#L52

roborourke:
Fixed in #177

PR #177, opened 2020-07-06
Turn on error display for local server

pdewouters:
Screenshot 2020-07-06 at 11 13 45

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -55,8 +55,6 @@ function bootstrap() {
 		'port' => getenv( 'REDIS_PORT' ),
 	];
 
-	ini_set( 'display_errors', 'on' );
-
 	if ( $config['tachyon'] ) {
 		define( 'TACHYON_URL', getenv( 'TACHYON_URL' ) );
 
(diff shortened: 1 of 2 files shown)


pdewouters:
@joehoyle are the acceptance criteria here outdated then? https://github.com/humanmade/altis-local-server/issues/123

joehoyle:
Ah thanks for linking @pdewouters -- could you link the original issues in subsequent PRs too to give extra context.

It looks like that acceptance criteria actually came from @roborourke! He edited my comment :D 

It does actually looked like we want to rmove display_errors from the web context per @rmccue 's response: 

> FWIW, it's already turned on in the local-server bootstrap file, but we should actually remove this so that errors aren't displayed in a HTML context:

I guess that's fine, though I'm not entirely sure why we don't want to show errors in the web context! In which case anyway, judging by the notes on #123 then this PR is good to go.

rmccue:
I think this should be set in the .ini rather than in wp-config FWIW

joehoyle:
@rmccue that will affect both web and cli though, which you comment specifically recommends against I think?

rmccue:
Yeah; I think local-server might need to distinguish the two SAPIs and separate the configuration, the same way that e.g. Ubuntu PHP does.

PR #189, opened 2020-07-17
Update Tachyon to 2.3.2

roborourke:
Fixes fallback to origin for gifs. Also adds webp detection support by checking `accept` header which makes media on local much much faster.


roborourke:
Here's a gif to test with

unnamed

PR #205, opened 2020-08-20
Fix incorrect regular expression to permit hyphenated-subdomains

kadamwhite:
The hyphen in .-_ is not treated as a literal hyphen unless it occurs at the end of the [] group. This can be demonstrated in the JS console:

```
'hyphenated-site'.match( /^[a-z.-_]+$/ ) // null
'hyphenated-site'.match( /^[a-z._-]+$/ ) // [ ...results ]
```

Reordering these characters allows hyphenated-subdomains to be matched.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -114,7 +114,7 @@ services:
       - "traefik.port=8080"
       - "traefik.protocol=https"
       - "traefik.docker.network=proxy"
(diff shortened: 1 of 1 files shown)


rmccue (approved):
I was thinking escaping via `\` would be better, but seems like there may be a YAML issue with that, so yolo

PR #208, opened 2020-08-26
Fix use of uploads directory on local server

roborourke:
After the change to the cloud module to ensure media paths matched the site's host name it broke local server's assumption that the uploads path was at `s3-<project>.altis.dev`.

This adds a 301 redirect on the nginx server to map to the correct URL on the S3 container.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -114,8 +114,9 @@ services:
       - "traefik.port=8080"
       - "traefik.protocol=https"
       - "traefik.docker.network=proxy"
-      - "traefik.uploads.frontend.redirect.regex=^https?://(.*?).altis.dev/uploads/(.*)"
-      - "traefik.uploads.frontend.redirect.replacement=https://s3-${COMPOSE_PROJECT_NAME:-default}.altis.dev/uploads/$$2"
+      # Redirect any requests to /uploads to the S3 container frontend.
+      - "traefik.uploads.frontend.redirect.regex=^https?://(?:.*?).altis.dev/uploads/(.*)"
+      - "traefik.uploads.frontend.redirect.replacement=https://s3-${COMPOSE_PROJECT_NAME:-default}.altis.dev/uploads/$$1"
       - "traefik.uploads.frontend.redirect.permanent=true"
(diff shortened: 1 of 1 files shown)


docker/docker-compose.yml
```diff
@@ -114,6 +114,9 @@ services:
       - "traefik.port=8080"
       - "traefik.protocol=https"
       - "traefik.docker.network=proxy"
+      - "traefik.uploads.frontend.redirect.regex=^https?://(.*?).altis.dev/uploads/(.*)"
+      - "traefik.uploads.frontend.redirect.replacement=https://s3-${COMPOSE_PROJECT_NAME:-default}.altis.dev/uploads/$$2"
```
rmccue, line 118: Should this use $$1 instead?

roborourke: no, I made that error first time round - there are 2 capturing brackets

rmccue: Sorry, I mean instead of the environment variable, not instead of the path replacement

roborourke: No, it's intentional. The S3 container URL will always be that

rmccue (approved):

docker/docker-compose.yml
```diff
@@ -114,6 +114,10 @@ services:
       - "traefik.port=8080"
       - "traefik.protocol=https"
       - "traefik.docker.network=proxy"
+      # Redirect any requests to /uploads to the S3 container frontend.
+      - "traefik.uploads.frontend.redirect.regex=^https?://(?:.*?).altis.dev/uploads/(.*)"
```
rmccue, line 118: Just to check, this won't cause an infinite redirect right??

roborourke: I did test it 😛  - the traefik rules apply first before the request gets to the container so `s3-*.altis.dev` requests are never handled by nginx.

joehoyle: WHen this says "redirect" does it mean HTTP redirect? If so, is this just used for legacy URLs? Just want to make sure we are not always issuing 301 redirects!

PR #211, opened 2020-08-27
Fix write permissions when running CLI commands

### LINKED ISSUE #183: PHP container doesn't have access to write to file system under Linux host
`ls -la` in `/usr/src/app` shows `1000` as the user (ID), should be www-data.

```
drwxrwxr-x   13 1000     1000          4096 Jul 14 14:36 .
drwxr-xr-x    1 root     root          4096 May  8 12:05 ..
-rw-rw-r--    1 1000     1000           280 Jul 13 15:27 .babelrc
drwxrwxr-x    2 1000     1000          4096 Jul 13 15:37 .bin
-rw-rw-r--    1 1000     1000           478 Jul 13 15:27 .build-script
```

joehoyle:
Pass the current user's UID when running on linux, as the mounted volumes are _directly_ accessed from the container. Unlike macOs, where the mounted volumes are is a Docker shared filesystem that correctly maps user permissions.

Fixes #183.

### DIFF
--- a/inc/composer/class-command.php
+++ b/inc/composer/class-command.php
@@ -378,10 +378,16 @@ EOT
 		$lines = exec( 'tput lines' );
 		$has_stdin = ! posix_isatty( STDIN );
 		$has_stdout = ! posix_isatty( STDOUT );
+		if ( php_uname( 's' ) === 'Linux' ) {
+			$user = '$UID';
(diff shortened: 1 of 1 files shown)


roborourke (approved):
Awesome, nice one @joehoyle !

PR #237, opened 2020-11-29
Allow "altis.dev" TLD to be configured

joehoyle:
With the "HTTPS required" natire of `.dev` domains, the fact that we
currently have to issue a fully signed wildcard cert (`.dev` doesn't
support self signed certs), and then the fact that we can't support
subdomain installs in local-server; I think the days of `altis.dev` are
probably numbered. This PR atleast makes this TLD fully configurable, to
something more `.local` or `.altis.local`.

In doing so, I also added an option to set `secure` to `true` / `false`,
so HTTP-only can then be supported with TLDs other than `.dev`.

At this point I don't think we need to publicly document this
neccesarily, but I think we are going to need to move away from this
hardcoding of altis.dev. If nothing else, this makes the code more
configurable in special use cases.

### DIFF
--- a/docker/conf/traefik.toml
+++ b/docker/conf/traefik.toml
@@ -153,7 +153,6 @@ address = ":8080"
 # Optional
 # Default: ""
 #
-domain = "altis.dev"
 
 # Expose containers by default in traefik
 #
--- a/inc/composer/class-command.php
+++ b/inc/composer/class-command.php
@@ -147,6 +147,7 @@ EOT
 		return [
 			'VOLUME' => getcwd(),
 			'COMPOSE_PROJECT_NAME' => $this->get_project_subdomain(),
+			'COMPOSE_PROJECT_TLD' => $this->get_project_tld(),
 			'PATH' => getenv( 'PATH' ),
 			'ES_MEM_LIMIT' => getenv( 'ES_MEM_LIMIT' ) ?: '1g',
 		];
@@ -233,7 +234,7 @@ EOT
 			$output->writeln( '<info>WP Password:</>	<comment>password</>' );
 		}
 
-		$site_url = 'https://' . $this->get_project_subdomain() . '.altis.dev/';
+		$site_url = $this->get_project_url();
 		$output->writeln( '<info>Startup completed.</>' );
 		$output->writeln( '<info>To access your site visit:</> <comment>' . $site_url . '</>' );
 
@@ -347,7 +348,7 @@ EOT
 	 * @return int
 	 */
 	protected function exec( InputInterface $input, OutputInterface $output, ?string $program = null ) {
-		$site_url = 'https://' . $this->get_project_subdomain() . '.altis.dev/';
+		$site_url = $this->get_project_url();
 		$options = $input->getArgument( 'options' );
 
 		$passed_url = false;
@@ -539,7 +540,7 @@ EOT;
 	 * @return void
 	 */
 	protected function generate_docker_compose() : void {
-		$docker_compose = new Docker_Compose_Generator( $this->get_project_subdomain(), getcwd() );
+		$docker_compose = new Docker_Compose_Generator( $this->get_project_subdomain(), getcwd(), $this->get_project_tld() );
 		// phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_read_file_put_contents
 		file_put_contents(
 			getcwd() . DIRECTORY_SEPARATOR . 'vendor' . DIRECTORY_SEPARATOR . 'docker-compose.yml',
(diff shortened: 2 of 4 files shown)


docker/conf/traefik.toml
```diff
@@ -153,7 +153,6 @@ address = ":8080"
 # Optional
 # Default: ""
 #
-domain = "altis.dev"
```
joehoyle, line 156: @roborourke I couldn't actually see where this was used. The docs say it's a default, do you know impact of removing this?

roborourke: We could maybe generate that file too. Honestly best off asking @nathanielks but if it all works without it then that’s ok

roborourke (changes requested):
Thanks @joehoyle!

There are a couple of little bugs in there that I've commented on inline, plus I also just merged in something that adds a 3rd parameter to the docker file generator so there's a conflict there. Should be easy resolve, I'm just passing in an array of extra 'stuff' now so you could add the TLD to that.

Are there any issues with the built in certificates being mounted in the traefik container when not using the `.altis.dev` TLD?

I think what's maybe missing here is a way to add your own certs to the Traefik proxy if you do change the TLD. How would I use something like `mkcert` to create self signed certs and use them with local server? That's something I'd like to see docs for and potentially a command that can automate the use of `mkcert`, or show installation instructions if missing.

Last thing is why not document this now? I don't think this is something we would backport (though could make a preview release of course). I'd like to avoid adding things without documenting them without at least creating a follow up issue to address it.

docker/conf/traefik.toml
nathanielks: AFAIK this won't impact anything. As you say, it's a default and I'm assuming the value would already be set by the `Host` header.

nathanielks: [Docs](https://doc.traefik.io/traefik/v1.7/configuration/backends/docker/) for posterity

inc/composer/class-docker-compose-generator.php
```diff
@@ -55,11 +55,11 @@ class Docker_Compose_Generator {
 	 * @param string $project_name The docker compose project name.
 	 * @param string $root_dir The project root directory.
 	 */
-	public function __construct( string $project_name, string $root_dir ) {
+	public function __construct( string $project_name, string $root_dir, string $tld ) {
```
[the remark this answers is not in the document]
joehoyle: As it happens, I think I can remove. We already pull the project config in this class anyway, so we can just read TLD from there

joehoyle:
> Are there any issues with the built in certificates being mounted in the traefik container when not using the .altis.dev TLD?

Yeah, so atm, you'd need to manually replace those in `vendor/altis/local-server` which is a hack. I was using `secure: false` myself to test / use this.

> I think what's maybe missing here is a way to add your own certs to the Traefik proxy if you do change the TLD. How would I use something like mkcert to create self signed certs and use them with local server? That's something I'd like to see docs for and potentially a command that can automate the use of mkcert, or show installation instructions if missing.

Yeah would love to use `mkcert` for exactly that. I don't know about how to update traefik to support multiple projects, all using their own certs. it might be that you can just list them all out and it will cleverly use the one matching the request host. There is https://github.com/traefik/traefik/pull/2233 which I was looking at. I've been following along with https://github.com/traefik/traefik/issues/5507#issuecomment-633622029 to support dynamic certs from labels, which would be ideal.

> Last thing is why not document this now? I don't think this is something we would backport (though could make a preview release of course). I'd like to avoid adding things without documenting them without at least creating a follow up issue to address it.

Specifically because we haven't solved SSL yet. I'd recommend on merging, and have it as an undocumented feature until we solve SSL too.

inc/composer/class-command.php
```diff
@@ -644,20 +645,59 @@ protected function minio_client( string $command ) {
...
+	/**
+	 * Get the root name to use for the project.
+	 *
+	 * @return string
+	 */
+	protected function get_project_tld() : string {
+		$config = $this->get_composer_config();
+
+		if ( isset( $config['tld'] ) ) {
+			$project_name = $config['tld'];
```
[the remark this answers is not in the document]
joehoyle: Hmm yeah I'd be for `die( $message )` to make it clear, though we might want to bubble that error yeah.

joehoyle:
@rmccue do you have anything to commit here in your trials with Codespaces. E.g. I think you accounted for no-told (`localhost`)

rmccue:
Yeah, work under way in https://github.com/humanmade/altis-local-server/tree/support-codespaces to support tld-less configurations, but Codespaces needs a few more things which I'll need to push up too.

roborourke:
Yeah looks like #341 has all the same stuff so we can close this in favour of that one. It'll need to be rebased on master or at least have the merge conflicts resolved.

PR #241, opened 2020-12-04
Elasticsearch version config option

roborourke:
Adds a config option to select versions 6.3, 6.8 or 7.10 of Elasticsearch that switches the containers used.

Using ES 7.10 allows for ARM support for both the ES and Kibana containers.

Fixes humanmade/product-dev#762


roborourke:
Need to audit all other tools that use ES currently and ensure that they're compatible. Known issues right now:

- `hits.hits.total` is `hits.hits.total.value` as of ES 7
  - Affects analytics audience size estimation feature

jazzsequence:
Is this still in progress?

roborourke:
Yep. It's mostly there but isn't a high priority right now

docs/elasticsearch.md
```diff
@@ -2,36 +2,68 @@
...
+}
+```
+
+The current available versions are:
+
+- 7.10
+- 6.8 (default)
+- 6.3
+
+You can also use the major version on its own to get the latest minor version, for example "6" will resolve to version "6.8".
```
rmccue, line 29: Should we note ARM is only compatible with 7 here?

shadyvb (approved):

docs/elasticsearch.md
roborourke: Yis

roborourke:
```suggestion
You can also use the major version on its own to get the latest minor version, for example "6" will resolve to version "6.8".

**Note**: If your device has an ARM chip you must use Elasticsearch 7 or higher.
```

PR #244, opened 2020-12-15
Upgrade to PHP 7.4 container

### LINKED ISSUE #243: Upgrade for PHP 7.4 container

roborourke:
As this update also includes XDebug v3 there are additional updates to support the updated config values and xdebug.mode setting.

Fixes #243

### DIFF
--- a/inc/composer/class-command.php
+++ b/inc/composer/class-command.php
@@ -108,7 +108,9 @@ EOT
 
 		// Collect args to pass to the docker compose file generator.
 		$docker_compose_args = [];
-		if ( $input->getOption( 'xdebug' ) ) {
+
+		// If Xdebug switch is passed add to docker compose args.
+		if ( $input->hasParameterOption( '--xdebug' ) ) {
 			$docker_compose_args['xdebug'] = $input->getOption( 'xdebug' );
 		}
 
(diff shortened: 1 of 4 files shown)


roborourke:
Needs a bit more testing as I'm having docker for mac issues after the upgrade.

roborourke:
This is good for review now.

inc/composer/class-command.php
```diff
@@ -68,7 +70,7 @@ protected function configure() {
 	import-uploads                Copies files from `content/uploads` to s3
 EOT
 			)
-			->addOption( 'xdebug' );
+			->addOption( 'xdebug', null, InputOption::VALUE_OPTIONAL, 'Start the server with XDebug', 'debug' );
```
jazzsequence, line 73: Shouldn't this use `Xdebug` instead of `XDebug`, too?

```suggestion
			->addOption( 'xdebug', null, InputOption::VALUE_OPTIONAL, 'Start the server with Xdebug', 'debug' );
```

jazzsequence (approved):
There's a bunch of linting issues here but the actual code looks fine. One minor comment since a lot of the docs switched `XDebug` to `Xdebug` but there's one new line of code where `XDebug` is being used/output.

inc/composer/class-command.php
roborourke: Ah nice one. I realised I'd been writing it wrong all this time!

jazzsequence: I had to look it up because I assumed it was XDebug, too 😜

PR #255, opened 2021-01-19
Enable XRay by default

### LINKED ISSUE #254: XRay container is off by default but Cloud module stills loads XRay code
Steps to reproduce:
1. Using the latest `master` branch spin a copy of local server
2. Make some requests
3. See XRay request errors in Query Monitor output

The Local Server config sets XRAy to not be set up by default however it doesn't account for the cloud module settings. This should be based off the cloud module setting for xray or at the very least it should not be off by default so that developers can consciously decide to turn it off for both the cloud module and local server together.

Acceptance criteria:

- [ ] Re-enable XRay by default

roborourke:
This was causing problem with the Cloud module's default setting fro XRay so we should avoid errors with a vanilla install. Devs will need to switch off both the cloud module setting and the local server setting to fully turn off xray locally.

Fixes #254

### DIFF
--- a/inc/composer/class-docker-compose-generator.php
+++ b/inc/composer/class-docker-compose-generator.php
@@ -647,7 +647,7 @@ class Docker_Compose_Generator {
 			'cavalcade' => true,
 			'elasticsearch' => true,
 			'kibana' => true,
-			'xray' => false,
+			'xray' => true,
 		];
 
 		return array_merge( $defaults, $config );


jazzsequence (changes requested):
This change is fine. However, before closing out #254 -- and given that it's indicated in the notes -- I would request that there be documentation added as part of #254 that explains how to disable Xray locally (and possibly why). Since the `fixed` keyword is used in the PR, merging this will auto-close that issue and I feel like there should be a separate, documentation component to this, hence requesting changes.

roborourke:
@jazzsequence we haven't documented any of the new config options for local server and the containers you can turn off yet because the feature isn't really complete. The other modules that would be affected by turning off say the ES container don't do any checks against the Local Server config so I'm not sure we're quite there yet.

Are you asking for a new issue to be created for the docs or for all the docs to be added to this PR? I'd rather do them all together when we're ready to. I'll create a new issue for that so that this change doesn't need to be held up.

roborourke:
Created https://github.com/humanmade/altis-local-server/issues/257 as a follow up to document the options.

jazzsequence (approved):
> Are you asking for a new issue to be created for the docs or for all the docs to be added to this PR?

I was specifically asking for documentation about turning Xray off, specifically, since it's mentioned in the PR description ("Devs will need to switch off both the cloud module setting and the local server setting to fully turn off xray locally.") -- if devs are expected to know a thing, the thing should be documented.

However, if the feature is as yet unannounced, and there's more associated documentation needed to cover all the aspects, and [we have an issue for _that_](https://github.com/humanmade/altis-local-server/issues/257), that's enough for me to have the ticket and toggle the default back to true.

PR #260, opened 2021-01-26
remove WP_DEBUG environment variables to prevent logs from not being readable via docker

nathanielks:


nathanielks:
I have a love/hate with `hub` sometimes. I didn't mean to file this.

PR #261, opened 2021-01-26
Disable debug log in favor of default error log

nathanielks:
With WP_DEBUG enabled, calls to `error_log()` are being sent to `content/debug.log` instead of the default error log. This is confusing, as I would expect logs to be visible via `composer serve logs php`. This restores that functionality.


joehoyle:
I think this is from https://github.com/humanmade/product-dev/blob/master/.config/debug.php#L9, which I think we might be better off changing in the skeleton repo rather than here?

rmccue:
Agreed, that seems like a better place to resolve that.

nathanielks:
Aha, indeed indeed! I'll file over there instead 👍

PR #263, opened 2021-02-08
Use traefik to pass s3 404s to nginx for fallback handling

kadamwhite:
For a while now we've been looking for a way to implement something akin to [Chassis `media_fallback`](https://github.com/chassis/media_fallback) within local server. This is a useful and important ability because it would let Altis site developers get their local up and running quickly using only a database dump, without also having to clone down potentially hundreds upon hundreds of media files.

Setting a local to use a production s3 bucket isn't ideal, because then developers cannot upload files to the local environment for testing. I feel that the most flexible setup would be to permit normal local file handling but then to conditionally redirect media 404's to nginx, so that we can use `nginx-additions` to catch those 404s and attempt to redirect them to a production file.

Adding an error fallback to the `tachyon` service didn't seem to do the trick, but after logging the request flow in traefik I was able to get 404s to consistently show up in nginx by setting these rules on the minio container. Once I got that working, then with the lines below in `nginx-additions.conf`

```nginx
location ~* ^/tachyon.+\.(jpe?g|gif|png|webp).*$ {
	try_files $uri @productionImages;
}

location @productionImages {
	rewrite ^/tachyon/(.*)$ https://{tachyon service domain}/{app name}/uploads/$1 break;
}
```
seems to do what I want; but without changes here in local-server, this behavior won't be possible.

I'd welcome feedback on the approach and my solution to it.

### DIFF
--- a/docs/media.md
+++ b/docs/media.md
@@ -1,0 +1,28 @@
+# Local Media
+
+Local Server supports the [Altis Media module](https://www.altis-dxp.com/resources/developer-docs/media/) by replicating the media handling capabilities of a deployed Altis environment. A new Local Server setup will support [Dynamic Images](https://www.altis-dxp.com/resources/developer-docs/media/dynamic-images/) and other media features, and all images you upload will be stored in a local container on your development machine.
+
+When using Local Server to work on an existing Altis site, you may need to pull down a database backup for local testing. Downloading uploaded media along with this database backup may be prohibitively time consuming, or take too much local disk space. Local Server can be configured to request these images directly from an Altis cloud environment.
+
+You can add a file `.config/nginx-additions-media.conf` to your Altis project which instructs Local Server to redirect image requests to a remote server if they were not accessible locally. This preserves your ability to upload and edit local images, while allowing any image file not present in your environment to be loaded from a remote bucket instead.
+
+Assuming that your Altis project uses the name `myproject` and you have confirmed that your environment uses the `us-east-1.tchyn.io` Tachyon URL, your `nginx-additions-media.conf` could look like this:
+
+<pre><code>
+# Listen for Tachyon 404s, and try redirecting those requests to the development S3 bucket.
+location ~* ^/tachyon.+\.(jpe?g|gif|png|webp).*$ {
+	try_files $uri @imageFallback;
+}
+
+location @imageFallback {
+	rewrite ^/tachyon/(.*)$ https://us-east-1.tchyn.io/myproject-development/uploads/$1 break;
+}
+</code></pre>
+
+Alternatively, for read-only image support you can define the constants `S3_UPLOADS_BUCKET`, `S3_UPLOADS_KEY`, `S3_UPLOADS_BUCKET_URL`, `S3_UPLOADS_REGION`, `S3_UPLOADS_SECRET`, and `TACHYON_URL`, to point your Local Server instance at a remote Tachyon instance. You will not be able to upload or edit images, but your local site should display as expected using the remote image server.
+
+Open a support ticket for assistance in determining the specific values and URLs necessary to configure either of these image fallback options on your site.
+
+## Refreshing Local Media
+
+If you chose to download images to your `uploads/` folder instead of using one of the above remote fallback approaches, you may need to signal Local Server to synchronize your local files with the image service. To run a local image sync, use the command `composer server import-uploads`.
--- a/docs/media.md
+++ b/docs/media.md
@@ -15,7 +15,7 @@ location ~* ^/tachyon.+\.(jpe?g|gif|png|webp).*$ {
 }
 
 location @imageFallback {
-	rewrite ^/tachyon/(.*)$ https://us-east-1.tchyn.io/myproject-development/uploads/$1 break;
+	rewrite ^/tachyon/(.*)$ https://myproject.altis.cloud/tachyon/$1 break;
 }
 </code></pre>
 
--- a/docs/media.md
+++ b/docs/media.md
@@ -1,6 +1,6 @@
 # Local Media
 
-Local Server supports the [Altis Media module](https://www.altis-dxp.com/resources/developer-docs/media/) by replicating the media handling capabilities of a deployed Altis environment. A new Local Server setup will support [Dynamic Images](https://www.altis-dxp.com/resources/developer-docs/media/dynamic-images/) and other media features, and all images you upload will be stored in a local container on your development machine.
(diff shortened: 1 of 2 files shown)


kadamwhite:
The main optimization I can think of over what I've got in place in that nginx config code snippet would be to somehow restrict the nginx location to only kick in if you're on an `altis.dev` URL. However, I'm not sufficiently versed in nginx config to know how to make that happen, if possible.

roborourke:
Ideally it would be useful to be able to generate the nginx config as well to make this easier to adopt for those who want it but we only include additional nginx config from the `/usr/src/app/.config` directory. That's not necessarily a problem given that on cloud environments the tachyon and uploads requests never hit the web server. The generated file could read from values entered in the `composer.json` config for this module perhaps.

I'd say that's optional though.

At a minimum this PR needs some accompanying documentation capturing the example config, and how to determine the substituted values you need to add before it's ready but I'm happy with the changes otherwise.

One thing that just occurred to me is how this behaves when running multiple instances of Local Server, my suspicion is that the name of the `backend` should be specific to the current project so Traefik knows which project's nginx container to route to.

kadamwhite:
> Ideally it would be useful to be able to generate the nginx config as well to make this easier to adopt for those who want it

@roborourke I disagree, I think that leaving this flexible for now until we're able to see how it gets used in reality is better; but +1 on the documentation and server de-dupe'ing.

roborourke:
I’m ok with that, as long as there’s docs. My reasoning is really coming from the Altis design principles, that we should automate when we can, but as you say, let’s see how popular it is and how else it might get used

docs/media.md
```diff
@@ -0,0 +1,28 @@
...
+
+When using Local Server to work on an existing Altis site, you may need to pull down a database backup for local testing. Downloading uploaded media along with this database backup may be prohibitively time consuming, or take too much local disk space. Local Server can be configured to request these images directly from an Altis cloud environment.
+
+You can add a file `.config/nginx-additions-media.conf` to your Altis project which instructs Local Server to redirect image requests to a remote server if they were not accessible locally. This preserves your ability to upload and edit local images, while allowing any image file not present in your environment to be loaded from a remote bucket instead.
+
+Assuming that your Altis project uses the name `myproject` and you have confirmed that your environment uses the `us-east-1.tchyn.io` Tachyon URL, your `nginx-additions-media.conf` could look like this:
```
roborourke, line 9: I think that's a legacy tachyon endpoint, it can be accessed via `https://example.com/tachyon/` now so we can simplify this part of the docs a bit:
```suggestion
Assuming that your Altis project uses the name `myproject` for Local Server and your environment uses the domain name `myproject.altis.cloud`, your `nginx-additions-media.conf` could look like this:
```

docs/media.md
```diff
@@ -0,0 +1,28 @@
...
+
+<pre><code>
+# Listen for Tachyon 404s, and try redirecting those requests to the development S3 bucket.
+location ~* ^/tachyon.+\.(jpe?g|gif|png|webp).*$ {
+	try_files $uri @imageFallback;
+}
+
+location @imageFallback {
+	rewrite ^/tachyon/(.*)$ https://us-east-1.tchyn.io/myproject-development/uploads/$1 break;
```
roborourke, line 18: Follow up to the previous comment:
```suggestion
	rewrite ^/tachyon/(.*)$ https://myproject.altis.cloud/tachyon/$1 break;
```

docs/media.md
```diff
@@ -0,0 +1,28 @@
...
+}
+</code></pre>
+
+Alternatively, for read-only image support you can define the constants `S3_UPLOADS_BUCKET`, `S3_UPLOADS_KEY`, `S3_UPLOADS_BUCKET_URL`, `S3_UPLOADS_REGION`, `S3_UPLOADS_SECRET`, and `TACHYON_URL`, to point your Local Server instance at a remote Tachyon instance. You will not be able to upload or edit images, but your local site should display as expected using the remote image server.
+
+Open a support ticket for assistance in determining the specific values and URLs necessary to configure either of these image fallback options on your site.
```
roborourke, line 24: I think let's leave this part out, those values could be found by logging on to the sandbox server and running `printenv` but we don't define or provide `S3_UPLOADS_KEY` or `S3_UPLOADS_SECRET` as environment variables, those credentials are handled by an instance profile. I don't think we really want to get in the habit of generating credentials for folks to use locally as they have to be rotated every so often, it's just more busy work for us.

Maybe lets just change it to:
```suggestion
Your production Tachyon URL will be whatever the primary domain name for the stack is plus the path `/tachyon/` so it may not always look like the example above.
```

inc/composer/class-docker-compose-generator.php
```diff
@@ -222,7 +222,7 @@ protected function get_service_nginx() : array {
 					'traefik.protocol=https',
 					'traefik.docker.network=proxy',
 					"traefik.frontend.rule=HostRegexp:{$this->hostname},{subdomain:[a-z.-_]+}.{$this->hostname}",
-					'traefik.backend=nginx',
+					"traefik.backend=nginx-{$this->hostname}",
```
roborourke: Just to confirm this still works ok? I'm not sure what character restrictions and things there are for these values and whether that's a concern.

roborourke (changes requested):
Looking good, just a couple of changes suggested

docs/media.md
```diff
@@ -0,0 +1,28 @@
+# Local Media
+
+Local Server supports the [Altis Media module](https://www.altis-dxp.com/resources/developer-docs/media/) by replicating the media handling capabilities of a deployed Altis environment. A new Local Server setup will support [Dynamic Images](https://www.altis-dxp.com/resources/developer-docs/media/dynamic-images/) and other media features, and all images you upload will be stored in a local container on your development machine.
```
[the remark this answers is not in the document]
roborourke: Oh for reference this is a shorthand we have built in for handling internal links between doc sets. More info here https://www.altis-dxp.com/resources/developer-docs/documentation/linking/#cross-linking

inc/composer/class-docker-compose-generator.php
kadamwhite: It didn't, which is why I pushed another commit to use project_name :slightly_smiling_face:

docs/media.md
kadamwhite: Works for me

roborourke (approved):

roborourke:
This will be officially released in v7 so you'll need to explicitly point to this commit or `dev-master` to use it in a project right now. We could look at making a preview release if preferred due to potential disruption.

PR #264, opened 2021-02-09
Expand logging instructions to explain underlying docker commands

kadamwhite:
While working on #263 I was a bit stuck before @roborourke explained how to directly access Traefik logs via docker. Once I understood how that logging command works, I found it more useful to directly follow the logs using `docker` commands because it removed some of the boilerplate text from the output and let me see more data at once. I think it would be useful to be explicit about how the logging command works under the hood to de-obfuscate this for developers who may need to dive deeper.


roborourke (approved):

roborourke:
Cheers @kadamwhite

PR #267, opened 2021-02-18
Improve file sharing performance with Mutagen

### LINKED ISSUE #204: Shared folder performance slow on macOs hosts
It's fairly well documented that the shared folder / mounted volumes are slow on docker-for-mac. See https://www.padok.fr/en/blog/docker-sync-osx-mounted-filesystem, https://docs.docker.com/docker-for-mac/osxfs-caching/, https://docs.docker.com/docker-for-mac/osxfs/.

We currently use the `:delegated` option in the `/usr/src/app` mount -- this is the best performance option Docker has. I've observed, even with that, file system latency is circa 10x slower than native access. The nature of Composer + WordPress applications is a _lot_ of filesystem access, as such the _majority_ of the load time on a fresh Altis install homepage load is filesystem access. A filesystem syscall is around 1microsocond natively, and about 10microseconds under the shared mount.

One possible solution to this is http://docker-sync.io/, which has been created to fix this limitation of Docker on Mac. I switched my local install to using it, and I achieved total load times of around 220ms (as quick as 180ms), compared to a typical 900ms when using macos docker mounts.

I don't exactly know the best solution here -- we might be able to support docker-sync when it is installed and available (and could be another reason to push ahead with https://github.com/humanmade/altis-local-server/pull/159

roborourke:
Note this is only valid for Mac and Linux currently.

It creates a top level volume for docker-compose which is shared by the services that need to access. In theory this means we are syncing once rather than the same stuff 3 times. In local development I have seen built files fail to sync for long periods of time requiring restarting the PHP container. It is painful to say the least.

Fixes #204


nathanielks (approved):
heyyy, nice 👍 smart man 👏 👏 👏

roborourke:
Would be awesome if someone else could also verify this improvement.

joehoyle:
@rmccue what do you mean by syncing? The file-sharing is a true share, with a custom file system implementation on osx, so it's not copied / synced to 3 places. With this change, I'm not sure how that exactly works though, does this cause an actual copy / duplication of files to the docker volume? (the current solution does not do that).

Also, I'd recommend we put this behind a flag for now -- I've got a lot of experience with NFS locally have small races / issues with things like incremental file updates, where perhaps 1-2% of file changes have a long delay to show up on the NFS share. I'd recommend using this for a while with opt-in to rule that out.

roborourke:
@joehoyle flag sounds sensible.

What I don't understand though is if it's a true share why it can take a very long time for files to become available on the containers. Surely that could only be the case if there's some sort of synchronisation going on.

Your work on resolving this issue by using docker in docker made me think that a single volume in the docker VM accessed by the PHP and nginx containers would achieve something similar. I saw an immediate improvement with this change personally but I'd like to get a few more guinea pigs to try it out first.

joehoyle:
> What I don't understand though is if it's a true share why it can take a very long time for files to become available on the containers. Surely that could only be the case if there's some sort of synchronisation going on.

I don't know the internal implementation exactly, but even things like _reads_ are 10x slower than the host, so it's defo not just syncing it to a place in the docker volume. 

> Your work on resolving this issue by using docker in docker made me think that a single volume in the docker VM accessed by the PHP and nginx containers would achieve something similar. I saw an immediate improvement with this change personally but I'd like to get a few more guinea pigs to try it out first.

I think that makes sense, I think for most cases with Altis, it's actually better to have the source files _in the container_ and then sync them _out_. I don't think docker has a great way to do that, and I think that's how some 3rd party tools work. In terms of needing to read quickly (and run PHP quickly), it's all happening inside the docker container.

roborourke:
What we need is middle-out volume technology! Cheers for the extra info though. I’ll run this locally for a bit and see how it goes

roborourke:
@joehoyle changed tack to using `docker-sync` was pretty easy to set up after all. Provided you've installed `docker-sync` you can now run `composer server start --docker-sync`

roborourke:
Turns out `docker-sync` is a bit broken. Maybe it's just the combination of Docker Desktop and OSX version...

Took another approach using Mutagen but it relies on the Mutagen Beta channel for the `mutagen compose` command. It can be installed using `brew install mutagen-io/mutagen/mutagen-beta` and then running `composer serve --mutagen` with this branch.

Mostly works - strange thing is directories created by the containers are synced to the host but files and file edits are not... Sort of ok for most work but it should be working.

roborourke:
Fixed the remaining issue with Mutagen! Needed to only set the file and folder permissions on the beta target.

It is _fast_.

This is good for a proper review now.

inc/composer/class-command.php
```diff
@@ -175,6 +200,21 @@ protected function get_env() : array {
 	protected function start( InputInterface $input, OutputInterface $output ) {
 		$output->writeln( '<info>Starting...</>' );
 
+		if ( $input->hasParameterOption( '--docker-sync' ) || $this->has_docker_sync_file() ) {
```
joehoyle, line 203: Hmm I'm not sure if we want `|| $this->has_docker_sync_file()`. If I want to start without docker-sync, after I've tried it, how would I do that?

I'd recommend we read from the config for the value if you don't want someone to have to pass `--docker-sync` every time. (also, support for a local config file would help there!)

inc/composer/class-command.php
```diff
@@ -175,6 +200,21 @@ protected function get_env() : array {
...
+			$sync = new Process( 'docker-sync start', 'vendor' );
+			// Might need to download the docker-sync container so extend this beyond default 60s.
+			$sync->setTimeout( 1200 );
+			$sync->setTty( true );
+			$sync_failed = $sync->run( function ( $type, $buffer ) {
```
joehoyle, line 208: I guess $buffer is a string? Can we type hint them both?

inc/composer/class-command.php
```diff
@@ -553,6 +614,33 @@ protected function generate_docker_compose( array $args = [] ) : void {
 			getcwd() . DIRECTORY_SEPARATOR . 'vendor' . DIRECTORY_SEPARATOR . 'docker-compose.yml',
 			$docker_compose->get_yaml()
 		);
+
+		// Put docker-sync.yml config in place if requested.
+		if ( $args['docker-sync'] ) {
```
joehoyle, line 619: I think this should probably go into another method

inc/composer/class-command.php
```diff
@@ -671,7 +759,75 @@ protected function get_project_subdomain() : string {
...
+		if ( $is_installed !== null ) {
+			return $is_installed;
+		}
+		if ( self::is_linux() || self::is_macos() ) {
+			$is_installed = ! empty( shell_exec( 'which mutagen' ) );
+		} else {
+			$is_installed = ! empty( shell_exec( 'where mutagen' ) );
+		}
+		// Create stub mutagen config file.
+		if ( $is_installed ) {
```
joehoyle, line 800: This is a `is_*` function, it should not have side effects like this IMO.

inc/composer/class-docker-compose-generator.php
```diff
@@ -624,6 +625,44 @@ public function get_array() : array {
 				'socket' => null,
 			],
 		];
+
+		// Handle docker-sync volume according to args.
+		if ( $this->args['docker-sync'] ?? false ) {
```
joehoyle, line 630: `! empty()` is the general de-facto way to do this I think

inc/composer/class-docker-compose-generator.php
```diff
@@ -624,6 +625,44 @@ public function get_array() : array {
...
+								'defaultFileMode' => '0664',
+								'defaultDirectoryMode' => '0775',
+							],
+						],
+						'mode' => 'two-way-resolved',
+						'ignore' => [
+							'paths' => [
+								'.DS_Store',
+								'.git',
+								'node_modules',
```
joehoyle, line 657: This technically differs from what we have now, which I think _could_ lead to some unexpected behavior.

roborourke: Hm, it's incredibly unlikely. Git files and node_modules are only interacted with on the host machine, both node and git are not installed on any of the containers were sharing files with.

This is really one of the main advantages and the reason for doing this in the first place. Take the product-dev repo for example. If I'm working on say analytics and smart media and have `node_modules` installed for both I could have over 200k files in my project, doing this means it only needs to worry about sharing something closer to 40k files so it's much more efficient.

inc/composer/class-command.php
roborourke: Yeah I could move that elsewhere 👍

inc/composer/class-command.php
roborourke: Maybe. We haven't type hinted it anywhere else in the file but could do.

inc/composer/class-command.php
roborourke: The files for both mutagen and docker-sync are removed after a destroy so right now you'd have to either destroy and re-up or remove the file manually. I could make it so the flag accepts a value like `--docker-sync=off` to remove it.

This is needed because the `docker-compose.yml` file has to be generated every time to apply settings. Once you start the server in this way it needs to stay consistent when running other commands so that the differences can be factored in eg. when restarting, stopping or destroying the server.

roborourke:
Cheers for the review @joehoyle. Thinking I might just remove `docker-sync` as an option anyway because it straight up didn't work at all. Mutagen seems much more robust.

inc/composer/class-docker-compose-generator.php
joehoyle: Well, in the current implementation, the amount of files you are sharing doesn't impact anything, because it's a filesystem wrapper. Unless the guest machine tries to read them, there is no overhead to move files. That might not be the same with your two supported methods of syncing files however.

I wonder if we could at least support a config option for it -- I know written code that can interact with node_modules (with things like v8js bridges, manifests etc).

joehoyle:
@roborourke sounds good, simpler = better!

roborourke:
@joehoyle ok that should do it - now supports passing `--mutagen=off` to switch it off. I've updated the docs and removed `docker-sync` support.

docs/README.md
```diff
@@ -50,8 +56,9 @@ The subdomain used for the project can be configured via the `modules.local-serv
 
 ## Available Commands
 
-* `composer server start [--xdebug]` - Starts the containers.
-  * If the `--xdebug` option is passed the PHP container will have XDebug enabled. To switch off XDebug run this command again without the `--xdebug` option.
+* `composer server start [--xdebug=<mode>] [--mutagen=(on|off)] [--docker-sync]` - Starts the containers.
+  * `--xdebug` will enable Xdebug. To switch off Xdebug run this command again without the `--xdebug` option.
+  * `--mutagen=(on|off)` will toggle file sharing using [Mutagen](https://mutagen.io/) if installed. Note that the Mutagen Beta version is required.
```
joehoyle, line 61: This doesn't seems very cli idiomatic. If clean my vendor directory or do a fresh install (or upgrade) I'll need to re-add --mutagen, as the file will be gone from the be vendor dir. Therefor it seems the only logical think to do as a user would be to specify it all the time.

I think it's a lot for expected that passing the flag is ephemeral (just like I assume --xdebug is)-- like I say, I think the config is the right place for any persistent settings

roborourke: If you removed your `vendor` dir it'd remove the `docker-composer.yml` as well. Problem is there isn't really a good place to put files specific to the project that we know are ignored by git. Upgrading wouldn't remove the files if that helps.

This needs to be a flag because the config is a committed file, if one person wants to use it then everyone has to which we don't want.

The difference with the `--xdebug` flag is that even if you didn't specify it when running `composer server stop` or `composer server destroy` it would still be successfully cleaned up. It's not quite as simple when using Mutagen.

It might not be idiomatic but it seems like the best user experience to me. I don't see as being super different to something like `docker login` on the CLI, that changes the behaviour of other commands. I do take your point but I'm not sure what the best option is.

joehoyle: I think of we want to take the persistent via cli option, I'd expect it to be another command like "server set-filesysten mutagen" or something along those lines

roborourke: I see yeah. That option might make more sense for the xdebug setting too. Creating and checking for files is the easiest way to do that but perhaps not in the `vendor` directory as you mention. Are you aware of any best practices for storing persistent settings? Maybe in a `.altis` directory in the users home directory.

roborourke:
@joehoyle ok - updated again, using a persistent settings store and added `composer server set <key> <value>`. I'll handle changing Xdebug to use the settings in a separate update assuming this works for you.

joehoyle:
@roborourke been mulling this over -- I think what we have here is a parallel configuration system just for local-server. We already have a configuration system in Altis, so I don't think introducing a second one just for local-server is a good idea. As you pointed out, we can't current have local-only configs, or user-wide config. Ultimately I think it would be better to support those things in the main Altis config, rather than introduce something special here.

I think we could quite trivially add support in Altis Core for a local-only config that is `.gitignore`d by default. I know that's a bit of a tangent here though. Sorry for giving the run-around on different directions. However, I don't want to go by the sunk-cost fallacy and merge this if it's not the right direction. My preferred approach would be:

1. Don't add persistence of any kind in this PR -- `--mutagen` should be passed to `start` on every call. Allow it to also read from the Altis config.
2. Add support to Altis Core for a local-only config in a separate PR. I don't know if that needs an RFC or something.

roborourke:
Hehe. Nah, having gone through the steps it did seem odd to have that persistent config live somewhere hard to see and not be part of core for all modules. Ideally we would provide a single entrypoint for all modules to register commands through. Will come back to the issue of local configs another time. I'm just excited to get this through because the performance is really good!

Moved to the ephemeral option approach just like the `--xdebug` flag. The problem I was seeing is more or less resolved by just using the `mutagen compose` wrapper command for `docker-compose` if it's available. That will handle stopping, starting and cleaning up syncs depending on whether the `--mutagen` flag is set.

jazzsequence (approved):
Some notes for the documentation piece, nothing that would block merge. Tested and seems to work for me (I didn't encounter any errors) but not sure how to really stress test it. 

Approving but not merging if further discussion about integration is needed.

joehoyle (approved):
Looks good, nothing beyond @jazzsequence's comments, nice one!

docs/mutagen-file-sharing.md
```diff
@@ -0,0 +1,66 @@
...
+{
+	"extra": {
+		"altis": {
+			"modules": {
+				"local-server": {
+					"ignore-paths": [
+                        ".git",
+                        "node_modules",
+                        ".DS_Store"
+                    ]
```
roborourke, line 58:
```suggestion
					"ignore-paths": [
						".git",
						"node_modules",
						".DS_Store"
					]
```

PR #271, opened 2021-03-09
Add a `mysql:` link to the database connection information output

johnbillion:
If you use a local MySQL client that supports the `mysql:` protocol but not the Sequel Pro `.spf` file format, for example [TablePlus](https://tableplus.com/), having this link present in the database connection details output allows you to connect straight to the local server database in that client (eg. by cmd+clicking the link in the terminal).


roborourke (approved):
🔥 

I like it, cheers @johnbillion

PR #274, opened 2021-03-16
Introduce "exec" subcommand for "composer server db"

kadamwhite:
I'm finding that I miss the ability to execute arbitrary sql commands and output the results to a local file. Enhancing the `db` command to accept an `exec` subcommand, which can take a string query like you would use directly with `mysql -e`, makes this easy without having to go via `db info` and then use the `mysql` command directly.

### DIFF
--- a/inc/composer/class-command.php
+++ b/inc/composer/class-command.php
@@ -70,6 +70,7 @@ Database commands:
 	db                            Log into MySQL on the Database server
 	db sequel                     Generates an SPF file for Sequel Pro
 	db info                       Prints out Database connection details
+	db exec -- "<query>"          Run and output the result of a SQL query.
 View the logs
 	logs <service>                <service> can be php, nginx, db, s3, elasticsearch, xray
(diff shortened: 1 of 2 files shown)


inc/composer/class-command.php
```diff
@@ -549,6 +550,18 @@ protected function db( InputInterface $input, OutputInterface $output ) {
...
+			case 'exec':
+				$query = $input->getArgument( 'options' )[1] ?? null;
+				if ( empty( $query ) ) {
+					$output->writeln( '<error>No query specified: pass a query via `db exec -- "sql query..."`</error>' );
+					break;
+				}
+				if ( substr( $query, -1 ) !== ';' ) {
+					$query = "$query;";
+				}
+				// phpcs:ignore WordPress.WP.CapitalPDangit.Misspelled
```
roborourke, line 563: lol

roborourke (approved):

kadamwhite:
Woo thank you!

PR #275, opened 2021-03-16
Clarify the Xdebug docs

johnbillion:
I was confused when I saw Xdebug in the list of extensions. This clarifies that the extension is always installed but not always active.


roborourke (approved):
Thanks John, sorry for the confusion. I did it this way since the update to Xdebug v3 allowing for the "off" mode. Makes switching it on and off a bit quicker.

PR #276, opened 2021-03-17
Require PHP 7.4 for Altis v7+

roborourke:


jazzsequence (approved):
We should make sure that the docs for this release explicitly say that we no longer support PHP 7.2 and if you need 7.2 support, you need to stick to v6.

roborourke:
Yeah for sure, we can tackle the upgrade guide first thing next sprint.

PR #290, opened 2021-05-21
Allow stopping specific services

joehoyle:
In some cases people may want to stop a given service (like cavalcade).


roborourke (changes requested):
Needs the markdown docs updating too and iff appropriate something for the troubleshooting guide as to when this is useful

roborourke (approved):

PR #291, opened 2021-05-28
Update the docs for the restart command

johnbillion:
The `restart` subcommand optionally accepts the name of a service such as `nginx`. This updates the docs to reflect that.

Ref: https://github.com/humanmade/altis-local-server/blob/master/inc/composer/class-command.php#L351-L356


roborourke (approved):
Thanks @johnbillion !

PR #293, opened 2021-06-08
require patch releases of 5.3 only

jazzsequence:
5.4 breaks compatibility with PHP < 8.1

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
 	"require": {
 		"php": ">=7.1",
 		"composer-plugin-api": "^1.1 || ^2.0",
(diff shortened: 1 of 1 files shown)


jazzsequence:
@roborourke should this and humanmade/altis-local-chassis#137 be backported as far as we support since this breaks pre-PHP 8 environments?

roborourke:
Yeah back to v4 :+1:

roborourke (approved):

jazzsequence:
These v4 and v5 backports are failing because we don't actually require `symfony/yaml` in either of those versions. No backport beyond v6 is needed.

PR #294, opened 2021-06-09
Allow to customize php image from ENV

ivankristianto:
Since we upgrade to use PHP 7.4 as the runtime on production. But we are still on Altis v5 which by default using PHP 7.2 from the docker image `humanmade/altis-local-server-php:3.2.0`

This PR only target `altis/local-server` version 5.x only.

To use new PHP image we can run with:
`PHP_IMAGE=humanmade/altis-local-server-php:4.0.0 composer server start`

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -13,7 +13,7 @@ x-php: &php
       condition: service_started
     mailhog:
       condition: service_started
-  image: humanmade/altis-local-server-php:3.2.0
+  image: ${PHP_IMAGE:-humanmade/altis-local-server-php:3.2.0}
   links:
     - "db:db-read-replica"
     - "s3:s3.localhost"
--- a/inc/composer/class-command.php
+++ b/inc/composer/class-command.php
@@ -146,6 +146,7 @@ EOT
 			'COMPOSE_PROJECT_NAME' => $this->get_project_subdomain(),
 			'PATH' => getenv( 'PATH' ),
 			'ES_MEM_LIMIT' => getenv( 'ES_MEM_LIMIT' ) ?: '1g',
+			'PHP_IMAGE' => getenv( 'PHP_IMAGE' ) ?: 'humanmade/altis-local-server-php:3.2.0',
 		];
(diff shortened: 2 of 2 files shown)


roborourke:
I think it might be better to provide this functionality for future versions too but through the config. In the meantime for older Altis versions I believe you could upgrade local server to v7 despite still being on Altis v5, I'm 90% certain they're compatible.

ivankristianto:
Thanks @roborourke 
I haven't thought about we can upgrade altis local server to latest version. I guess the way we version Altis Core and Altis Local Server, my first assumption it should match.
I will try to update SC codebase to use Altis Local server v7.x
It should work too in my opinion, because all the services doesn't need env variable that shared from Altis Core.

ivankristianto:
Tried to upgrade Altis local-server v7.0.0 on RB local env. Got this when I tried wp cli:
```
bash-5.0$ wp option get siteurl
Error: Site '.altis.dev/' not found. Verify DOMAIN_CURRENT_SITE matches an existing site or use `--url=<url>` to override.
```

roborourke:
What were the steps you took? Did you run `composer server start` again?

That specific error might just require you to set the `--url` option.

ivankristianto:
These lines cause the issue I reported when I tried the local-server v7.0.1.
https://github.com/humanmade/altis-local-server/blob/master/inc/namespace.php#L36-L38
```
$_SERVER['HTTP_HOST'] = getenv( 'COMPOSE_PROJECT_NAME' ) . '.altis.dev';
```

This env doesn't exist in docker php environments. It only exist when running composer.
I can send patch in different PR later.

ivankristianto:
PR opened here: https://github.com/humanmade/altis-local-server/pull/298
once that fixed, we can close this one, or do you think we can still use this PR for older release @roborourke to allow to modify the php image from the env/compose config?

roborourke:
Closing this PR, I think we need an issue to refine what we want to achieve in future. It will likely need a mapping of PHP versions to container versions.

PR #297, opened 2021-06-18
Use `docker compose` if available

### LINKED ISSUE #289: `docker-compose` moved to `docker compose` and shows notices
The latest version of the Docker CLI now incorporates `docker-compose` so ideally we need to check for the version of Docker and modify the commands accordingly.

roborourke:
Fixes #289


jazzsequence (approved):
👍

PR #304, opened 2021-07-19
Support for Linux on WSL

roborourke:
Linux on WSL can still use the `host.docker.internal` host name to resolve the Docker VM, so we need to account for that too when switching between the Docker Engine IP used in Linux vs Linux on WSL.

### DIFF
--- a/inc/composer/class-command.php
+++ b/inc/composer/class-command.php
@@ -727,6 +727,24 @@ EOT;
 	}
 
 	/**
+	 * Check if the current host is Windows.
+	 *
+	 * @return boolean
+	 */
+	public static function is_windows() : bool {
+		return php_uname( 's' ) === 'Windows';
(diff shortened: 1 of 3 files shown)


inc/composer/class-command.php
```diff
@@ -726,6 +726,24 @@ public static function is_macos() : bool {
...
+		return php_uname( 's' ) === 'Windows';
+	}
+
+	/**
+	 * Check if the current host is WSL.
+	 *
+	 * @return boolean
+	 */
+	public static function is_wsl() : bool {
+		return ! empty( $_ENV['WSL_INTEROP'] );
```
rmccue, line 744:
```suggestion
		return getenv( 'WSL_INTEROP' ) !== false;
```

rmccue (approved):

PR #310, opened 2021-07-22
Update external proxy network config

### LINKED ISSUE #309: Fix `network proxy: network.external.name is deprecated in favor of network.name` warning
Not sure when this was deprecated precisely but we should update the docker compose config to correct it - may require checking the docker version so needs a bit of exploration first.

roborourke:
Fixes #309


roborourke:
Took a while but found that this configuration format has been supported for a long time so we don't need to do any checking for the docker version. That said I don't think we need to backport beyond v6 right now as it's a deprecation. Folks should be updating before long.

shadyvb (approved):

PR #329, opened 2021-08-04
Fix docker command switching logic

roborourke:
Return code was 0 on linux for missing docker subcommand, this checks the actual output instead.


jazzsequence (approved):
👍

PR #328, opened 2021-08-04
Align output of stop with start

tfrommen:
Currently, the `composer server stop` command is not using TTY. This means that I would see this in my terminal:

```
$ composer server stop
Stopping...
Stopping my-project_nginx_1           ... 
Stopping my-project_cavalcade_1       ...
Stopping my-project_s3-sync-to-host_1 ...
Stopping my-project_php_1             ...
Stopping my-project_kibana_1          ...
Stopping my-project_s3_1              ...
Stopping my-project_elasticsearch_1   ...
Stopping my-project_xray_1            ...
Stopping my-project_mailhog_1         ...
Stopping my-project_tachyon_1         ...
Stopping my-project_cognito_1         ...
Stopping my-project_redis_1           ...
Stopping my-project_pinpoint_1        ...
Stopping my-project_db_1              ...
Stopping my-project_s3-sync-to-host_1 ... done
Stopping my-project_xray_1            ... done
Stopping my-project_cavalcade_1       ... done
Stopping my-project_cognito_1         ... done
Stopping my-project_pinpoint_1        ... done
Stopping my-project_nginx_1           ... done
Stopping my-project_php_1             ... done
Stopping my-project_redis_1           ... done
Stopping my-project_mailhog_1         ... done
Stopping my-project_db_1              ... done
Stopping my-project_s3_1              ... done
Stopping my-project_tachyon_1         ... done
Stopping my-project_kibana_1          ... done
Stopping my-project_elasticsearch_1   ... done
Stopping docker_proxy_1 ... 
Stopping docker_proxy_1 ... done
Stopped.
```

This PR aligns the `stop` command with the `start` one, updating it to use TTY and no timeout.

This results in the following output, where each container is updating its own line:

```
$ composer server stop
Stopping...
Stopping my-project_nginx_1           ... done
Stopping my-project_cavalcade_1       ... done
Stopping my-project_s3-sync-to-host_1 ... done
Stopping my-project_php_1             ... done
Stopping my-project_kibana_1          ... done
Stopping my-project_s3_1              ... done
Stopping my-project_elasticsearch_1   ... done
Stopping my-project_xray_1            ... done
Stopping my-project_mailhog_1         ... done
Stopping my-project_tachyon_1         ... done
Stopping my-project_cognito_1         ... done
Stopping my-project_redis_1           ... done
Stopping my-project_pinpoint_1        ... done
Stopping my-project_db_1              ... done
Stopping docker_proxy_1 ... done
Stopped.
```

PR #332, opened 2021-08-16
Set TTY depending on execution context for all processes

roborourke:
Makes sure that output is properly formatted.

### DIFF
--- a/inc/composer/class-command.php
+++ b/inc/composer/class-command.php
@@ -199,7 +199,7 @@ EOT
 
 		$proxy = new Process( $this->get_compose_command( '-f proxy.yml up -d' ), 'vendor/altis/local-server/docker' );
 		$proxy->setTimeout( 0 );
-		$proxy->setTty( true );
+		$proxy->setPty( true );
 		$proxy_failed = $proxy->run( function ( $type, $buffer ) {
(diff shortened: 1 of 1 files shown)


roborourke:
Here's the follow on @tfrommen!

rmccue:
These should be set based on whether the current process is a TTY, no?

roborourke:
The start command has had this for ages so we're bringing them in line with that https://github.com/humanmade/altis-local-server/pull/328

We only use tty detection for the `exec` command right now.

rmccue:
Seems like they should consistently do so though; otherwise, something like `composer server logs php | tee some.log | tail -f` will likely break.

roborourke:
Using `$process->setPty( true )` ([pseudo terminal](https://symfony.com/doc/current/components/process.html#using-tty-and-pty-modes)) works, Symfony figures out whether the output is being piped to something else or not.

tfrommen (approved):
**_Colors!_** 😍

rmccue:
I am thinking more like `$process->setTty( posix_isatty( STDOUT ) )`; passing the code through. Setting pty won't pass through interactive prompts if those would have occurred (iirc)

PR #340, opened 2021-09-17
Use max ES container memory limit for Xmx setting

roborourke:
A number of analytics queries fail without this increase with an error like the following (truncated):

`{"error":{"root_cause":[{"type":"circuit_breaking_exception","reason":"[parent] Data too large, data for [<reused_arrays>] would be [510575952/486.9mb], which is larger than the limit of [510027366/486.3mb], real usage: [510575872/486.9mb], new bytes reserved: [80/80b], usages [request=464680/453.7kb, fielddata=54107/52.8kb, in_flight_requests=5162/5kb, accounting=3157104/3mb]","bytes_wanted":510575952,"bytes_limit":510027366,"durability":"PERMANENT"},{"type":"circuit_breaking_exception","reason":"[parent] Data too large, data for [<reused_arrays>] would be [510575952/486.9mb], which is larger than the limit of [510027366/486.3mb], real usage: [510575872/486.9mb], new bytes reserved: [80/80b], usages [request=200256/195.5kb, fielddata=54107/52.8kb, in_flight_requests=5162/5kb, accounting=3157104/3mb]","bytes_wanted":510575952,"bytes_limit":510027366,"durability":"PERMANENT"},{"type":"circuit_breaking_exception","reason":"[parent] Data too large, data for [<agg [endpoint.Demographic.Locale]>] would be [510580992/486.9mb], which is larger than the limit of [510027366/486.3mb], real usage: [510575872/486.9mb], new bytes reserved: [5120/5kb], usages [request=73088/71.3kb, fielddata=54693/53.4kb, in_flight_requests=5162/5kb, accounting=3157104/3mb]","bytes_wanted":510580992,"bytes_limit":510027366,"durability":"PERMANENT"},{"type":"circuit_breaking_exception","reason":"[parent] Data too large, data for [<reused_`

We might be able to modify the queries a bit to work with small ES instances, or just generally make the queries less heavy but it will require a good deal of time and effort.

### DIFF
--- a/inc/composer/class-command.php
+++ b/inc/composer/class-command.php
@@ -182,7 +182,7 @@ EOT
 			'DOCKER_CLIENT_TIMEOUT' => 120,
 			'COMPOSE_HTTP_TIMEOUT' => 120,
 			'PATH' => getenv( 'PATH' ),
-			'ES_MEM_LIMIT' => getenv( 'ES_MEM_LIMIT' ) ?: '2g',
+			'ES_MEM_LIMIT' => getenv( 'ES_MEM_LIMIT' ) ?: '1g',
 			'HOME' => getenv( 'HOME' ),
 		];
 	}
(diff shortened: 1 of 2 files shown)


roborourke:
Not super keen on this but I'm not sure what else to do short term.

joehoyle (approved):

PR #341, opened 2021-09-19
Support Codespaces

rmccue:
Early draft of just getting things working. Built on #40.

One of the bigger changes here is changing how Minio works. Rather than using the automatic bucket domain building, this changes it to be more in line with the other tools with a static domain. Additionally, it disables the TLS validation, since Minio has the wrong certificate (slash, self-signed? dunno) for these purposes. (nb, I haven’t updated this for the latest changes which added the Traefik routing; that work probably wants to go back into #40 too)

With these changes, it’s zero-configuration to work on Codespaces. Just start a Codespace with an Altis project, and `composer install && composer server start`, and the linked domain will just work. That said, you can also use [a simplified dev container](https://github.com/humanmade/platform-test/pull/89) for a bit speedier setup.

This also works if you use Codespaces with VS Code desktop, but this instead uses `localhost` as the host, which is a bit annoying. Needs to be disabled somehow? I think there’s a VS Code setting for it idk

Needs more testing, and I am probably bad at code, so there might be cleaner ways to do some of the stuff. Also needs docs (and maybe we get a blog post up for it?)

(I’ve set the humanmade org to have $100 worth of credits, since we’re now out of the free period for Codespaces.)


roborourke:
Not super into the `secure` switch here but I guess browser APIs that require https should work on `localhost`. If that's needed for zero config set up then all good.

joehoyle:
> Not super into the secure switch here but I guess browser APIs that require https should work on localhost. If that's needed for zero config set up then all good.

Yeah that was from my PR, it was basically a way to support other domains etc until we get a way to generate the SSL certs etc.

joehoyle (approved):

roborourke:
It feels like a good candidate for packaging up the functionality, e.g. in a WP plugin using `set_url_scheme()` would have the desired effect. Fewer ternaries cluttering up the place.

@rmccue can you test this with the recent changes to the Minio frontends?

joehoyle:
@roborourke im not sure I follow-- how would a wp plugin change all the local server configuration to run services as http rather than https?

roborourke:
Sorry I just meant the `set_url_scheme()` function provided by WordPress could be useful here.

joehoyle:
@rmccue anything a blocker here now? is there hope to get this in v9?

roborourke:
Linter checks are failing still, and I think it needs to rebased on master or have master merged into it.

inc/composer/class-docker-compose-generator.php
```diff
@@ -129,16 +129,16 @@ protected function get_php_reusable() : array {
 				'ELASTICSEARCH_HOST' => 'elasticsearch',
 				'ELASTICSEARCH_PORT' => 9200,
 				'AWS_XRAY_DAEMON_HOST' => 'xray',
-				'S3_UPLOADS_ENDPOINT' => "https://{$this->tld}/",
+				'S3_UPLOADS_ENDPOINT' => 'http' . ( $this->args['secure'] ? 's' : '' ) . "://s3-{$this->hostname}.localhost/",
```
roborourke, line 132: For readability lets add a helper method to read the `secure` arg and modify the URL e.g.:

```php
$this->set_url_scheme( "http://s3-{$this->hostname}.localhost/" )
```

inc/composer/class-docker-compose-generator.php
```diff
@@ -63,12 +63,12 @@ class Docker_Compose_Generator {
 	 * @param string $root_dir The project root directory.
 	 * @param array $args An optional array of arguments to modify the behaviour of the generator.
```
roborourke, line 64:
```suggestion
	 * @param string $tld The TLD to use for the Local Environment, e.g. altis.dev.
 	 * @param array $args An optional array of arguments to modify the behaviour of the generator.
```

inc/composer/class-command.php
```diff
@@ -719,6 +779,11 @@ public static function is_macos() : bool {
 		return php_uname( 's' ) === 'Darwin';
 	}
 
+	public function is_using_codespaces() : bool {
```
roborourke, line 782: Missing doc block

load.php
```diff
@@ -15,6 +15,7 @@
 		's3' => true,
 		'tachyon' => true,
 		'analytics' => true,
+		'codespaces_integration' => true,
```
roborourke, line 18: I don't really understand the purpose of this config option.

shadyvb:
Refreshed rebased and extended at https://github.com/humanmade/altis-local-server/pull/466 , addressed feedback, closing this one down, but keeping the branch for comparison and referencing.

shadyvb:
Reopening after separating concerns between SSL and Codespaces support.

roborourke (approved):
Not tested myself yet but looks good - just one comment stating a todo to check on, guessing it just needs removing.

Good stuff!

inc/composer/class-docker-compose-generator.php
```diff
@@ -153,17 +163,17 @@ protected function get_php_reusable() : array {
 				'ELASTICSEARCH_HOST' => 'elasticsearch',
 				'ELASTICSEARCH_PORT' => 9200,
 				'AWS_XRAY_DAEMON_HOST' => 'xray',
-				'S3_UPLOADS_ENDPOINT' => "https://s3-{$this->hostname}/s3-{$this->project_name}/",
+				'S3_UPLOADS_ENDPOINT' => Command::set_url_scheme( "https://s3-{$this->hostname}/s3-{$this->project_name}/" ), // TODO .localhost.
```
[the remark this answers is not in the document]
shadyvb: Nope! Nice catch.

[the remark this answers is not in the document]
shadyvb:
```suggestion
				'S3_UPLOADS_ENDPOINT' => Command::set_url_scheme( "https://s3-{$this->hostname}/s3-{$this->project_name}/" ),
```

PR #252, opened 2021-01-04
Support for Arm64

joehoyle:

PR #343, opened 2021-09-19
Support for ARM64

ntwb:
Follow up to #252

This PR was created using Altis v8 with `local-server` release candidate `8.1.0-rc` [here](https://github.com/humanmade/altis-local-server/releases/tag/8.1.0-rc)

Notes: 
- All images are native ARM64, no containers run under emulation
- All images except for Mailhog are using the original Docker base image
- Minio ~12 months ago bumped MC to v10 when multi-arch images began being built, hence the v10 config bump
- I've used `latest` for Docker images here for simplicity, pinning to specific releases would be required for release
- I'm using `mysql/mysql-server:8.0`, rather than MariaDB, solely because I have a client project where I'm using this daily, I think the suggestion has been to prefer MariaDB over MySQL v8?

### Testing

From the dashboard -> tools menu e.g. https://subdomain.altis.dev/wp-admin/tools.php
image

- ❌ S3 Browser - https://s3-subdomain.altis.dev/minio
  - The following warning is displayed when opening the above URL:
    - `NoSuchKeyThe specified key does not exist.minios3-subdomain/minious-east-xxxxxxxxxxx`
- ✅ Kibana - https://subdomain.altis.dev/kibana/
  - Works as expected and opens Kibana at https://subdomain.altis.dev/kibana/app/home#/
- ✅ Mailhog - https://subdomain.altis.dev/mailhog/
  - Works as expected and opens Mailhog at https://subdomain.altis.dev/mailhog

### DIFF
--- a/inc/composer/class-docker-compose-generator.php
+++ b/inc/composer/class-docker-compose-generator.php
@@ -91,7 +91,7 @@ class Docker_Compose_Generator {
 					'condition' => 'service_started',
 				],
 			],
-			'image' => 'humanmade/altis-local-server-php:edge',
+			'image' => 'humanmade/altis-local-server-php:4.2.0-dev',
 			'links' => [
 				'db:db-read-replica',
 				's3:s3.localhost',
@@ -200,7 +200,7 @@ class Docker_Compose_Generator {
 	protected function get_service_nginx() : array {
 		return [
 			'nginx' => [
-				'image' => 'humanmade/altis-local-server-nginx:edge',
+				'image' => 'humanmade/altis-local-server-nginx:3.4.0',
 				'networks' => [
 					'proxy',
 					'default',
@@ -256,7 +256,7 @@ class Docker_Compose_Generator {
 	protected function get_service_db() : array {
 		return [
 			'db' => [
-				'image' => 'mysql/mysql-server:8.0',
+				'image' => 'biarms/mysql:5.7',
 				'volumes' => [
 					'db-data:/var/lib/mysql',
 				],
@@ -416,7 +416,7 @@ class Docker_Compose_Generator {
 	protected function get_service_s3() : array {
 		return [
 			's3' => [
-				'image' => 'minio/minio:latest',
+				'image' => 'minio/minio:RELEASE.2021-09-18T18-09-59Z',
 				'volumes' => [
 					's3:/data:rw',
 				],
@@ -455,7 +455,7 @@ class Docker_Compose_Generator {
 				],
 			],
 			's3-sync-to-host' => [
-				'image' => 'minio/mc:latest',
+				'image' => 'minio/mc:RELEASE.2021-09-02T09-21-27Z',
 				'restart' => 'unless-stopped',
 				'depends_on' => [
 					's3',
@@ -480,7 +480,7 @@ class Docker_Compose_Generator {
 	protected function get_service_tachyon() : array {
 		return [
 			'tachyon' => [
-				'image' => 'humanmade/tachyon:edge',
+				'image' => 'humanmade/tachyon:2.4.0',
 				'ports' => [
 					'8080',
 				],
@@ -513,7 +513,7 @@ class Docker_Compose_Generator {
 	protected function get_service_mailhog() : array {
 		return [
 			'mailhog' => [
-				'image' => 'cd2team/mailhog:latest',
+				'image' => 'cd2team/mailhog:1632011321',
 				'ports' => [
 					'8025',
 					'1025',
@@ -551,7 +551,7 @@ class Docker_Compose_Generator {
 					'default',
 				],
 				'restart' => 'unless-stopped',
-				'image' => 'humanmade/local-cognito:edge',
+				'image' => 'humanmade/local-cognito:1.1.0',
 				'labels' => [
 					'traefik.port=3000',
 					'traefik.protocol=http',
@@ -568,7 +568,7 @@ class Docker_Compose_Generator {
 					'default',
 				],
 				'restart' => 'unless-stopped',
-				'image' => 'humanmade/local-pinpoint:edge',
+				'image' => 'humanmade/local-pinpoint:1.3.0',
 				'labels' => [
 					'traefik.port=3000',
 					'traefik.protocol=http',
@@ -590,7 +590,7 @@ class Docker_Compose_Generator {
 	protected function get_service_xray() : array {
 		return [
 			'xray' => [
-				'image' => 'amazon/aws-xray-daemon:latest',
+				'image' => 'amazon/aws-xray-daemon:3.3.3',
 				'ports' => [
 					'2000',
 				],
--- a/inc/composer/class-docker-compose-generator.php
+++ b/inc/composer/class-docker-compose-generator.php
@@ -91,7 +91,7 @@ class Docker_Compose_Generator {
 					'condition' => 'service_started',
 				],
 			],
-			'image' => 'humanmade/altis-local-server-php:4.0.0-dev',
+			'image' => 'humanmade/altis-local-server-php:edge',
 			'links' => [
 				'db:db-read-replica',
 				's3:s3.localhost',
@@ -200,7 +200,7 @@ class Docker_Compose_Generator {
 	protected function get_service_nginx() : array {
 		return [
 			'nginx' => [
-				'image' => 'humanmade/altis-local-server-nginx:3.3.0',
+				'image' => 'humanmade/altis-local-server-nginx:edge',
 				'networks' => [
 					'proxy',
 					'default',
@@ -256,7 +256,7 @@ class Docker_Compose_Generator {
 	protected function get_service_db() : array {
 		return [
 			'db' => [
-				'image' => 'mysql:5.7',
+				'image' => 'mysql/mysql-server:8.0',
 				'volumes' => [
 					'db-data:/var/lib/mysql',
 				],
@@ -416,7 +416,7 @@ class Docker_Compose_Generator {
 	protected function get_service_s3() : array {
 		return [
 			's3' => [
-				'image' => 'minio/minio:RELEASE.2020-03-19T21-49-00Z',
+				'image' => 'minio/minio:latest',
 				'volumes' => [
 					's3:/data:rw',
 				],
@@ -455,7 +455,7 @@ class Docker_Compose_Generator {
 				],
 			],
 			's3-sync-to-host' => [
-				'image' => 'minio/mc:RELEASE.2020-03-14T01-23-37Z',
+				'image' => 'minio/mc:latest',
 				'restart' => 'unless-stopped',
 				'depends_on' => [
 					's3',
@@ -480,7 +480,7 @@ class Docker_Compose_Generator {
 	protected function get_service_tachyon() : array {
 		return [
 			'tachyon' => [
-				'image' => 'humanmade/tachyon:2.3.2',
+				'image' => 'humanmade/tachyon:edge',
 				'ports' => [
 					'8080',
 				],
@@ -513,7 +513,7 @@ class Docker_Compose_Generator {
 	protected function get_service_mailhog() : array {
 		return [
 			'mailhog' => [
-				'image' => 'mailhog/mailhog:latest',
+				'image' => 'cd2team/mailhog:latest',
 				'ports' => [
 					'8025',
 					'1025',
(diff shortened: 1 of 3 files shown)


joehoyle:
> I'm using mysql/mysql-server:8.0, rather than MariaDB, solely because I have a client project where I'm using this daily, I think the suggestion has been to prefer MariaDB over MySQL v8?

I think this is due to compatibility, the assumption being that Maria would be closer to MySQL 5.7 in terms of compat. We will need to switch Altis Cloud to MySQL 8 _at some point_ . I don't believe AWS Aurora MySQL supports 8 yet, so I don't think we'd be switching for some time though.

roborourke:
That minio error is kinda telling: `exist.minios3-subdomain/minious-east-xxxxxxxxxxx`

I think it'll work with a slash on the end eg. https://s3-subdomain.altis.dev/minio/ - if the admin tools menu link is missing that then we should add it. Maybe it's some change in minio 10. Let me know if that sorts it anyway. Maybe some other Minio admin related stuff has changed.

roborourke:
I'll go through an make tagged releases for anything referencing `latest` here unless they already exist.

The MySQL 8 issue seems to be the major hurdle here. Local databases will need to be recreated to use it iirc.

We could try this image in the interim https://hub.docker.com/r/biarms/mysql - not official but has a lot of pulls / usage.

inc/composer/class-docker-compose-generator.php
```diff
@@ -416,7 +416,7 @@ protected function get_service_kibana() : array {
 	protected function get_service_s3() : array {
 		return [
 			's3' => [
-				'image' => 'minio/minio:RELEASE.2020-03-19T21-49-00Z',
+				'image' => 'minio/minio:latest',
```
roborourke, line 419:
```suggestion
				'image' => 'minio/minio:RELEASE.2021-09-18T18-09-59Z',
```

inc/composer/class-docker-compose-generator.php
```diff
@@ -455,7 +455,7 @@ protected function get_service_s3() : array {
 				],
 			],
 			's3-sync-to-host' => [
-				'image' => 'minio/mc:RELEASE.2020-03-14T01-23-37Z',
+				'image' => 'minio/mc:latest',
```
roborourke, line 458:
```suggestion
				'image' => 'minio/mc:RELEASE.2021-09-02T09-21-27Z',
```

inc/composer/class-docker-compose-generator.php
```diff
@@ -480,7 +480,7 @@ protected function get_service_s3() : array {
 	protected function get_service_tachyon() : array {
 		return [
 			'tachyon' => [
-				'image' => 'humanmade/tachyon:2.3.2',
+				'image' => 'humanmade/tachyon:edge',
```
roborourke, line 483:
```suggestion
				'image' => 'humanmade/tachyon:2.4.0',
```

inc/composer/class-docker-compose-generator.php
```diff
@@ -513,7 +513,7 @@ protected function get_service_tachyon() : array {
 	protected function get_service_mailhog() : array {
 		return [
 			'mailhog' => [
-				'image' => 'mailhog/mailhog:latest',
+				'image' => 'cd2team/mailhog:latest',
```
roborourke, line 516:
```suggestion
				'image' => 'cd2team/mailhog:1632011321',
```

inc/composer/class-docker-compose-generator.php
```diff
@@ -551,7 +551,7 @@ protected function get_service_analytics() : array {
 					'default',
 				],
 				'restart' => 'unless-stopped',
-				'image' => 'humanmade/local-cognito:1.0.0',
+				'image' => 'humanmade/local-cognito:edge',
```
roborourke, line 554:
```suggestion
				'image' => 'humanmade/local-cognito:1.1.0',
```

inc/composer/class-docker-compose-generator.php
```diff
@@ -568,7 +568,7 @@ protected function get_service_analytics() : array {
 					'default',
 				],
 				'restart' => 'unless-stopped',
-				'image' => 'humanmade/local-pinpoint:1.2.3',
+				'image' => 'humanmade/local-pinpoint:edge',
```
roborourke, line 571:
```suggestion
				'image' => 'humanmade/local-pinpoint:1.3.0',
```

inc/composer/class-docker-compose-generator.php
```diff
@@ -590,7 +590,7 @@ protected function get_service_analytics() : array {
 	protected function get_service_xray() : array {
 		return [
 			'xray' => [
-				'image' => 'amazon/aws-xray-daemon:3.0.1',
+				'image' => 'amazon/aws-xray-daemon:latest',
```
roborourke, line 593:
```suggestion
				'image' => 'amazon/aws-xray-daemon:3.3.3',
```

roborourke (changes requested):
I think this should cover it

ntwb:
> That minio error is kinda telling: `exist.minios3-subdomain/minious-east-xxxxxxxxxxx`
> 
> I think it'll work with a slash on the end eg. https://s3-subdomain.altis.dev/minio/ - if the admin tools menu link is missing that then we should add it. Maybe it's some change in minio 10. Let me know if that sorts it anyway. Maybe some other Minio admin related stuff has changed.

It did not, same error though there's now a `/` between minio and the AWS region:

> `NoSuchKeyThe specified key does not exist.minio/s3-subdomain/minio/us-east-xxxxxxxxxxx`

And correct, the tools menu item doesn't include the trailing slash...

roborourke:
> It did not, same error though there's now a / between minio and the AWS region:

Cheers for confirming, will need to dig into the config then. There's a bit of faffing around to get the URLs to play nicely.

rmccue:
n.b. I made some slight improvements to the Minio/AWS SDK configuration in #341 that might be worth looking at here; in particular, [changing the Minio hostname](https://github.com/humanmade/altis-local-server/pull/341/files#diff-6e1aa008277f9688efac7746b347e1e418cfc7b7e36f5bec209e48022d735022R417) and the SDK config to [avoid dynamically building hostnames](https://github.com/humanmade/altis-local-server/pull/341/files#diff-a0f318343a79085d556c9f6050c1261041075f5129f00e908fce122d89e02febR31).

joehoyle:
I noticed in https://github.com/humanmade/altis-local-server/blob/arm64-support-stephen/inc/composer/class-command.php#L706 too we reference _a_ version of minio, does that need to be updated?

inc/composer/class-docker-compose-generator.php
```diff
@@ -480,7 +480,7 @@ protected function get_service_s3() : array {
 	protected function get_service_tachyon() : array {
 		return [
 			'tachyon' => [
-				'image' => 'humanmade/tachyon:2.3.2',
+				'image' => 'humanmade/tachyon:2.4.0',
```
joehoyle, line 483: @roborourke this docker image does not seem to exist.

inc/composer/class-docker-compose-generator.php
```diff
@@ -91,7 +91,7 @@ protected function get_php_reusable() : array {
 					'condition' => 'service_started',
 				],
 			],
-			'image' => 'humanmade/altis-local-server-php:4.0.0-dev',
+			'image' => 'humanmade/altis-local-server-php:4.2.0-dev',
```
joehoyle, line 94: @roborourke this docker image doesn't seem to exist.

joehoyle (changes requested):

joehoyle:
Ok I can confirm I'm working, with all ARM containers (I did some manual fix-ups to the above missing images):

I wrote a quick bash script to verify (all `dev_` containers for example).

```
docker ps | grep dev_ | awk '{print $1}' | xargs -I _ docker inspect _ -f '{{.Config.Image}}' | xargs -I z docker image inspect z -f '{{.Architecture}}'
arm64
arm64
arm64
arm64
arm64
arm64
arm64
arm64
arm64
arm64
arm64
arm64
arm64
arm64
amd64
```

The amd64 one at the end is ```565b39f66056   mutagenio/sidecar:latest                           "/usr/bin/mutagen-si…"   35 minutes ago   Up 35 minutes                                                                                                                                   dev_mutagen_1``` I'm not actually sure why I'm getting that though, because I didn't think I have mutagen on. However, do we need to look for that.

inc/composer/class-docker-compose-generator.php
roborourke: 🤦‍♂️  I see we have a bit of mix of release tags prepended with `v` on this repo but without on Docker hub. Should have been `v2.4.0`

inc/composer/class-docker-compose-generator.php
roborourke: 🤔  strange, I'll look into this

roborourke: I thought it was still pushing 2 versions of the container, looks look Docker hub just has the dev version of the container now. Guess that's fine given we don't pull from there for prod.
```suggestion
			'image' => 'humanmade/altis-local-server-php:4.2.0',
```

inc/composer/class-docker-compose-generator.php
roborourke:
```suggestion
				'image' => 'humanmade/tachyon:v2.4.0',
```

roborourke:
Ah ok, it won't be used unless you pass the `--mutagen` flag.

What version of the mutagen beta do you have installed? It looks like multiarch builds should be available with the 0.12.0 beta https://hub.docker.com/r/mutagenio/sidecar/tags

roborourke (approved):

roborourke:
For what it's worth mutagen is still an experimental feature so I don't see that as a blocker for this.

ntwb:
Yeah, I've been using mutagen locally, unsure how to fully test, but nothing seems to break, latest arm64 links and instructions are in the nightly docs too

- https://docs.altis-dxp.com/local-server/mutagen-file-sharing/

inc/composer/class-docker-compose-generator.php
ntwb: I'm still using the `edge` builds, haven't had a chance to try the versions Rob tagged, but I see 2.4.0 [here](https://hub.docker.com/layers/humanmade/tachyon/v2.4.0/images/sha256-e43878b7dc1a6ae8cb8b17a80ac28fff9ec133f63b4ef8be5f50dfd8f55b0bcd?context=explore)

inc/composer/class-docker-compose-generator.php
ntwb: I also don't see `4.2.0-dev`, but `4.2.0` is [here](https://hub.docker.com/layers/humanmade/altis-local-server-php/4.2.0/images/sha256-3a6efd764fede1b7efc44667b2516dbb33445a39bd6d145cd8baf23a3f137e44?context=explore)

inc/composer/class-docker-compose-generator.php
```diff
@@ -91,7 +91,7 @@ protected function get_php_reusable() : array {
 					'condition' => 'service_started',
 				],
 			],
-			'image' => 'humanmade/altis-local-server-php:4.0.0-dev',
+			'image' => 'humanmade/altis-local-server-php:4.2.0',
```
joehoyle, line 94: @roborourke `-dev` IIRC is the dev container, not a "dev version" that has xdebug intsalled: https://github.com/humanmade/docker-wordpress-php/blob/master/Dockerfile-dev. Did we stop using `-dev` just for the tag (and it's built from `-dev`?)-- sorry I can't remember but wanted to check, as I think it might have even been me who made that change!

roborourke: Ah gotcha, yeah so the GitHub workflow uses the `Dockerfile-dev` file to build. It wasn't a dev container - it was always just used to reference the version that had Xdebug installed. I think that's fine personally as Xdebug 3 can be present and have minimal impact. Plus we don't use the builds on Docker hub on our infrastructure.

roborourke: I didn't make that particular change, I just didn't notice it earlier - I thought we still had the regular tag and tag + `-dev` with Xdebug being built and pushed up.

The `-dev` suffix was previously how we toggled between a container with Xdebug and one without, but we use the `XDEBUG_MODE` setting to switch it on or off now.

PR #345, opened 2021-09-23
Add import command

joehoyle:
This makes it easy to import a Altis Dashboard backup into the local server.

There's a few ways this could also be done:

- In HM CLI / Node etc
- In the cloud module
- In a shared place where chassis can make use of it.

All that is true, but I thought I'd start somewhere, as I had to continually import different backups so why not push it up!

### DIFF
--- a/inc/composer/class-command.php
+++ b/inc/composer/class-command.php
@@ -13,6 +13,9 @@
 namespace Altis\Local_Server\Composer;
 
 use Composer\Command\BaseCommand;
+use GuzzleHttp;
+use PharData;
+use Symfony\Component\Console\Helper\ProgressBar;
 use Symfony\Component\Console\Input\ArrayInput;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
@@ -75,6 +78,7 @@ View the logs
 	logs <service>                <service> can be php, nginx, db, s3, elasticsearch, xray
 Import files from content/uploads directly to s3:
 	import-uploads                Copies files from `content/uploads` to s3
+	import <export-url>           Import database and files from an Altis Cloud export URL.
 EOT
 			)
 			->addOption( 'xdebug', null, InputOption::VALUE_OPTIONAL, 'Start the server with Xdebug', 'debug' )
@@ -161,6 +165,8 @@ EOT
 			return $this->shell( $input, $output );
 		} elseif ( $subcommand === 'import-uploads' ) {
 			return $this->import_uploads( $input, $output );
+		} elseif ( $subcommand === 'import' ) {
+			return $this->import( $input, $output );
 		} elseif ( $subcommand === null ) {
 			// Default to start command.
 			return $this->start( $input, $output );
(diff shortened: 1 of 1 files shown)


joehoyle:
Has a couple todo's still, but I thought I'd get thoughts from @roborourke on the approach.

inc/composer/class-command.php
```diff
@@ -687,6 +693,192 @@ protected function import_uploads() {
...
+			// Flush cache
+			$flush_cache = $cli->run( new ArrayInput( [
+				'subcommand' => 'cli',
+				'options' => [
+					'cache',
+					'flush'
+				],
+			] ), $output );
+
+			$elasticpress_index = $cli->run( new ArrayInput( [
```
roborourke, line 870: We also have the `wp altis migrate` command now but that's technically not required, just wondering if we can trigger everything hooked onto the `multisite-install` command instead, without actually running that command!

roborourke (commented):
It looks like this could be wholesale lifted and shifted into the https://github.com/humanmade/altis-dev-tools-command repo given the use of `$this->getApplication()`. That command has a wrapper for running the exec command on either local server or local chassis with the `--chassis` flag.

So with this I would need to be able to get the export URL from the Altis dashboard - I don't think client devs have access to those necessarily so I'd be keen for us to solve that first before we roll this out in a stable release.

I'm happy with the approach on the whole, we don't have to wait for Altis/HM CLI, especially given that Altis/HM CLI could just run this command under the hood if we did abstract it.

Definitely good to get features to devs sooner than later though.

Does this deal with uploads too? Depending on the dev environment it would require running `composer server import-uploads` too to get them into the S3 container.

joehoyle:
> It looks like this could be wholesale lifted and shifted into the https://github.com/humanmade/altis-dev-tools-command repo given the use of $this->getApplication(). That command has a wrapper for running the exec command on either local server or local chassis with the --chassis flag.

Cool, I wasn't sure if `altis dev-tools import` implies enough that it's importing it to your local machine, but if we think that's not a concern, then moving it there would be fine.

> So with this I would need to be able to get the export URL from the Altis dashboard - I don't think client devs have access to those necessarily so I'd be keen for us to solve that first before we roll this out in a stable release.

All "User" and above get to read backups, see https://github.com/humanmade/vantage-backend/blob/staging/content/mu-plugins/hm-stack-api/inc/endpoints/class-backups-endpoint.php#L21. You just right click on the backup "export" and Copy URL

> Does this deal with uploads too? Depending on the dev environment it would require running composer server import-uploads too to get them into the S3 container.

Yes, I didn't do uploads yet, but needs to be added.

> We also have the wp altis migrate command now but that's technically not required, just wondering if we can trigger everything hooked onto the multisite-install command instead, without actually running that command!

Hmm yeah I'm not sure. If I import a product database, maybe I want to be able to test the `migrate` command for example? (so run it manually myself)

roborourke:
> Cool, I wasn't sure if altis dev-tools import implies enough that it's importing it to your local machine, but if we think that's not a concern, then moving it there would be fine.

That's a really good point. A terse command name is ideal though so not something I want to bikeshed, perhaps in future if there are more command line interactions with cloud it can go under a namespace like `composer dev-tools cloud import` or we add a new command package to get `composer cloud <subcommand>`

joehoyle:
Kinda related, it would be very nice to be able to get an export of the analytics data, and import that locally too. No idea how feasible that is. Perhaps in our backups we could also start putting the ES analytics data in the export... stretch goal n all, but would make it a lot easier to work with analytics stuff if you could pull it from production to local.

---

- [x] Implement uploads handling
- [ ] Add all cleanup todo's
- [ ] Move to appropriate command / chassis abstractions

roborourke:
There is the export endpoint but requires authentication, additionally that could add up to 1.5gb of data a day in some cases, with 90 indexes that’s masses of data, and potentially personally identifiable data too. Would be best to have some redaction tools in place for any local env commands

roborourke:
> Cool, I wasn't sure if `altis dev-tools import` implies enough that it's importing it to your local machine, but if we think that's not a concern, then moving it there would be fine.

Looking at this again - it is a good point. Could argue it either way but we're going to deprecate Local Chassis so that's not a blocker for this any more. This will be a nice short term helper until we can make time to work on Altisifying HM CLI.

I'll write up an issue so it can be picked up, updated and planned into a sprint.

roborourke:
@joehoyle think we can pick this back up again, Local Chassis is officially removed now

PR #348, opened 2021-10-04
Always restart Elastic container unless stopped

shadyvb:
Elastic container can easily shutdown because of memory issues, this ensures the container is restarted if that happens to avoid distrupting the developer workflow.

### DIFF
--- a/inc/composer/class-docker-compose-generator.php
+++ b/inc/composer/class-docker-compose-generator.php
@@ -311,6 +311,7 @@ class Docker_Compose_Generator {
 		return [
 			'elasticsearch' => [
 				'image' => $image,
+				'restart' => 'unless-stopped',
 				'ulimits' => [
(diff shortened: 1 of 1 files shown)


roborourke:
Will this solve the issue that causes the container to crash in the first place or obscure it?

If it's crashing during some process and then coming back up you might not realise the reason for data being out of sync, a command not working etc...

Ideally I'd love to be able to provide some sort of notification or log output if a container becomes unhealthy or crashes, perhaps via Query Monitor or even desktop notifications if possible. That way we might be able to better surface the troubleshooting guide or suggest solutions like increasing memory via `ES_MEM_LIMIT`, Docker desktop settings, etc...

shadyvb:
@roborourke when you say log an error message, do you mean to PHP's error log, or to the admin dashboard somewhere ?

PR #350, opened 2021-10-06
Fix container names for Docker Compose 1.29.2

### LINKED ISSUE #349: PHP Container not found error during start up
When starting Local Server an error occurs stating "PHP Container not found to run command". WP does not get installed and the site / application cannot be accessed.

roborourke:
Docker Compose just shipped a new version that generates container names differently using hyphens instead of underscores. We had a hardcoded check for the name `_php_1` in the exec command which failed with the new container name scheme.

Fixes #349

### DIFF
--- a/docker/proxy.yml
+++ b/docker/proxy.yml
@@ -4,6 +4,7 @@ services:
     networks:
       - proxy
     image: traefik:1.7
+    container_name: altis-proxy
     volumes:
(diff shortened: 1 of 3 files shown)


shadyvb (approved):

roborourke:
I checked and docker compose just renames existing containers so this will work fine. The error message I added should be enough to guide people. There’s no impact on volumes or stored data either, just the container names.

PR #362, opened 2021-10-06
Fix Minio Browser/Console for Minio v10

### LINKED ISSUE #346: Fix compatibility with Minio v10
The Minio browser appears not to be working with v10, which is required for the ARM64 multiarch builds.

Specifically: the Minio file browser is no longer accessible. When opening the file browser, it shows the results for a file by that name (this key does not exist) instead of the browser itself.

This could be from configuration changes in Minio or could be Traefik.

This is from humanmade/product-dev#596, however is not a blocker to shipping.

Two potential acceptance criteria here:
* [ ] File browser is accessible at the URL we link out to; **or**
* [ ] If not doable, link to Minio file browser is removed

Size: S

roborourke:
Fixes #346


roborourke:
Not urgent, figured this out while debugging the inability to upload reported earlier on Slack.

PR #404, opened 2022-01-21
Ensure to disable pseudo-tty when get db info

ivankristianto:
Without `-T`:
```
❯ composer server db info

[detail truncated]

[ErrorException]
  Undefined index: MYSQL_ROOT_PASSWORD
```

With `-T`:
```
❯ composer server db info
Root password:  [truncated]

Database:       [truncated]
User:           [truncated]
Password:       [truncated]

Host:           0.0.0.0
Port:           58388

Version:        5.7.35-1debian10
```

I'm currently on Altis v6 but i see this might be reproduce-able on v9.

### DIFF
--- a/inc/composer/class-command.php
+++ b/inc/composer/class-command.php
@@ -667,7 +667,7 @@ EOT;
 		$lines = exec( 'tput lines' );
 
 		$base_command = sprintf(
-			"$command_prefix %s exec -e COLUMNS=%d -e LINES=%d db",
+			"$command_prefix %s exec -e COLUMNS=%d -e LINES=%d -T db",
 			$this->get_compose_command(),
 			$columns,
 			$lines


roborourke:
Pretty sure this was fixed, you might need to update docker desktop though. Can you confirm your docker desktop and docker compose version?

roborourke:
Also check you’re on the latest patch version of this package

ivankristianto:
@roborourke 
These are my environments:
- MacOS 12.1
- Docker Desktop 4.4.2 (73305) 
- Docker Compose version v2.2.3
- Composer version 2.1.3
- PHP 7.4.25 (cli)
- Altis Local Server v6.0.7

And the `composer server db info` will throw error when not apply the `-T` flag.

PR #428, opened 2022-02-04
Support mounting tmp directory and using webgrind for profiling

roborourke:
Xdebug generates output files in the `/tmp` directory that can be read by KCacheGrind, QCacheGrind and WebGrind for profiling. This covers the bases that XHProf misses as it can be used to profile Cavalcade and CLI invocations.

This also uses the as yet unused `tmp` volume and shares it between the PHP containers meaning `composer server logs php` provides a combined error log from both cavalcade and PHP.

Lastly to support other Xdebug related output and `/tmp` related debugging the `--tmp` option will mount the `/tmp` directory to `.tmp` in the project root in order to provide easy access to files.


shadyvb (approved):
Nice one!

PR #433, opened 2022-03-02
Allow ssh as an alias for shell

tomjn:
This will save a massive amount of frustration and anger towards this package by allowing muscle memory from 10 years of vagrant to work, rather than continuously pressing backspace and retyping ssh as shell. It will also significantly cut down on the number of `--help` commands issued.

Packages should make engineers happy, not angry.

### DIFF
--- a/inc/composer/class-command.php
+++ b/inc/composer/class-command.php
@@ -39,7 +39,7 @@ class Command extends BaseCommand {
 			->setName( 'server' )
 			->setDescription( 'Altis Local Server' )
 			->setDefinition( [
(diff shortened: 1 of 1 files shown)


tfrommen:
Just wondering, has this not been documented on purpose? I mean, we can just silently allow for `ssh` to work, sure. But we might want to document this, both [in the terminal](https://github.com/humanmade/altis-local-server/blob/79d276eb182bd0c6b8f6234536cbde048b77fd85/inc/composer/class-command.php#L70-L71), as well as the [docs](https://github.com/humanmade/altis-local-server/blob/79d276eb182bd0c6b8f6234536cbde048b77fd85/docs/README.md#available-commands)...?

PR #438, opened 2022-03-30
Fix symfony console version to 5.x

roborourke:
Composer 2.3 allows for Symfony console v6 which has a breaking change for the `Process` object. This fixes the version to one that composer 2.3 allows but is still compatible with Altis.


mikelittle (approved):
Looks good to me

PR #441, opened 2022-03-30
Add support for Composer 2.3 version of Symfony Console

roborourke:
Composer 2.3 updates `symfony/console` to 5.4 which requires an array for the first argument. This checks the version and uses a proxy function to support whichever version is running.

I tried finding a few ways to properly find the installed version of `symfony/console` with composer itself but it's not straightforward at all. Additionally if you install `symfony/console` as a dependency of this package Composer doesn't use it so we have to rely on the composer version to know which symfony packages are in use.

### DIFF
--- a/inc/composer/class-command.php
+++ b/inc/composer/class-command.php
@@ -208,7 +208,7 @@ EOT
 	protected function start( InputInterface $input, OutputInterface $output ) {
 		$output->writeln( '<info>Starting...</>' );
 
-		$proxy = new Process( $this->get_compose_command( '-f proxy.yml up -d' ), 'vendor/altis/local-server/docker' );
+		$proxy = $this->process( $this->get_compose_command( '-f proxy.yml up -d' ), 'vendor/altis/local-server/docker' );
 		$proxy->setTimeout( 0 );
 		$proxy->setTty( posix_isatty( STDOUT ) );
 		$proxy_failed = $proxy->run( function ( $type, $buffer ) {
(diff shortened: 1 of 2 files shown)


roborourke:
Sorry to tag you @rmccue, will be helpful to try and resolve this overnight as it's gonna generate a lot of support requests atm.

tomjn:
related: https://blog.packagist.com/composer-2-3/ and https://github.com/composer/composer/pull/10435/files

swtiching to arrays is recommended but it's not intended to be necessary

inc/composer/class-command.php
```diff
@@ -781,6 +782,42 @@ protected function get_project_subdomain() : string {
...
+	 * @return string|null
+	 */
+	protected function get_composer_version() : ?string {
+		static $version;
+
+		if ( $version ) {
+			return $version;
+		}
+
+		preg_match( '/^Composer ([\d\.-]+)/', exec( 'composer --version' ), $composer_version );
```
rmccue, line 814: Could this use `Composer::getVersion()` instead since we're already running in Composer context?

roborourke: Wasn't aware that was a thing! Ta

inc/composer/class-command.php
```diff
@@ -12,6 +12,7 @@
 
 namespace Altis\Local_Server\Composer;
 
+use Composer\Composer;
```
rmccue, line 15: Lint

roborourke:
@tomjn we're using the Symfony Console component directly, specifically its `Process` object which does now require an array. `composer.lock` for Composer 2.3 itself specifies `symfony/console` 5.4, previously it has been locked to `2.8.52` for quite a while.

rmccue (approved):

PR #465, opened 2022-04-22
Add ssl command to download mkcert and generate ssl certificates

shadyvb:
ref https://github.com/humanmade/product-dev/issues/987


inc/composer/class-command.php
```diff
@@ -654,6 +661,179 @@ protected function db( InputInterface $input, OutputInterface $output ) {
...
+					case 'Darwin':
+						$binary_arch = $arch === 'x86_64' ? 'darwin-amd64' : 'darwin-arm64';
+						break;
+					# Linux
+					case 'Linux':
+						$binary_arch = $arch === 'amd64' ? 'linux-amd64' : 'linux-arm64';
+						break;
+					# Windows
+					case 'Windows':
+						$binary_arch = 'windows-amd64.exe';
```
roborourke, line 692: Will need to check what the situation with this and WSL is. If you can run .exe from WSL, or global commands from there we might be ok. @missjwo might be able to assist in testing.

inc/composer/class-command.php
```diff
@@ -654,6 +661,179 @@ protected function db( InputInterface $input, OutputInterface $output ) {
...
+				// Setup and accept the root certificate.
+				exec( "$mkcert -install", $dummy, $result );
+				if ( $result ) {
+					$output->writeln( "<error>Could not setup mkcert properly, try manually installing mkcert.</error>" );
+					return 1;
+				}
+
+				$output->writeln( "<info>mkcert root CA was installed and accepted successfully.</info>" );
+				return 0;
+				break;
```
roborourke, line 748: I'm not sure you need to both `return` and `break`. Not sure how PHP deals with it versus javascript.

inc/composer/class-command.php
```diff
@@ -654,6 +661,179 @@ protected function db( InputInterface $input, OutputInterface $output ) {
...
+					return 1;
+				}
+
+				$binary = "mkcert-$mkcert_version-$binary_arch";
+				$mkcert = "vendor/mkcert";
+
+				// Check if mkcert is installed globally already, bail if so.
+				$version = trim( shell_exec( 'mkcert -version' ) );
+				if ( $version ) {
+					$output->writeln( "<error>mkcert $version is installed globally already</error>" );
```
roborourke, line 711: Does it have a way to check if the root CA is already set up? Bit edge case but there's a chance someone might not have done that yet. I think that can probably be covered by an FAQ entry though.

inc/composer/class-command.php
```diff
@@ -654,6 +661,179 @@ protected function db( InputInterface $input, OutputInterface $output ) {
...
+				}
+
+				// Check if mkcert is installed locally already, bail if so.
+				$version = trim( shell_exec( "$mkcert -version" ) );
+				if ( $version ) {
+					$output->writeln( "<error>mkcert $version is installed locally already</error>" );
+					return 1;
+				}
+
+				exec( "curl -o $mkcert -L https://github.com/FiloSottile/mkcert/releases/download/$mkcert_version/$binary", $dummy, $result );
```
roborourke, line 722: Windows might need `wget`, unless WSL has `curl` by default.

inc/composer/class-command.php
```diff
@@ -654,6 +661,179 @@ protected function db( InputInterface $input, OutputInterface $output ) {
...
+				if ( $result ) {
+					$output->writeln( "<error>Could not download mkcert binary, try using sudo or manually installing mkcert.</error>" );
```
roborourke, line 724: Lets add another line with a link to the `mkcert` repo or releases page as above.

inc/composer/class-command.php
```diff
@@ -654,6 +661,179 @@ protected function db( InputInterface $input, OutputInterface $output ) {
...
+					return 1;
+				}
+
+				// TODO figure out how to programmatically detect the domains to use
+				$domains = $input->getArgument( 'options' )[1] ?? '*.altis.dev';
+
+				exec( "$mkcert -cert-file vendor/ssl-cert.pem -key-file vendor/ssl-key.pem $domains", $dummy, $result );
+
+				if ( $result ) {
+					$output->writeln( "<error>Could not generate certificates! Try generating them manually using mkcert.</error>" );
```
roborourke, line 762: Lets provide an example command so users know the intended cert file and key file locations. Also will it generate these files per domain or all in one file? Just wondering about overwriting. Likely not a problem but something to consider.

inc/composer/class-command.php
```diff
@@ -654,6 +661,179 @@ protected function db( InputInterface $input, OutputInterface $output ) {
...
+				break;
+			case 'generate':
+				$mkcert = $this->get_mkcert_binary();
+				if ( ! $mkcert ) {
+					$output->writeln( "<error>mkcert is not installed, run 'composer server ssl install' or install mkcert manually first.</error>" );
+					return 1;
+				}
+
+				// TODO figure out how to programmatically detect the domains to use
+				$domains = $input->getArgument( 'options' )[1] ?? '*.altis.dev';
```
roborourke, line 757: Can I provide a space separated list of domains still?

roborourke: Regarding programmatically detecting the domains to use let's just use the `composer.json` config, maybe we can have the `tld` option accept an array of TLDs and we generate certs for each by default. Would need to merge that configured list with whatever might be passed to this command.

inc/composer/class-command.php
```diff
@@ -654,6 +661,179 @@ protected function db( InputInterface $input, OutputInterface $output ) {
...
+
+				exec( "$mkcert -cert-file vendor/ssl-cert.pem -key-file vendor/ssl-key.pem $domains", $dummy, $result );
+
+				if ( $result ) {
+					$output->writeln( "<error>Could not generate certificates! Try generating them manually using mkcert.</error>" );
+					return 1;
+				}
+
+				$output->writeln( "<info>Generated SSL certificate successfully to vendor/ssl-cert.pem and key to vendor/ssl-key.pem.</info>" );
```
roborourke, line 766: You can always break up longer lines, this might be easier to view as a list. Not a requirement, just command readability is a good place for sweating the small details.

inc/composer/class-command.php
```diff
@@ -654,6 +661,179 @@ protected function db( InputInterface $input, OutputInterface $output ) {
...
+					$output->writeln( "<error>$exec_output</error>" );
+					return 1;
+				} else {
+					$output->writeln( $exec_output );
+				}
+
+				break;
+
+			case '':
+				$mkcert = $this->get_mkcert_binary();
```
roborourke, line 789: There's a fair bit of duplication, I was thinking the `get_mkcert_binary()` can take the `$output` object as an argument but that's overloading its purpose a bit. Perhaps the error message string can just be stored as a variable and reused.

inc/composer/class-command.php
```diff
@@ -654,6 +661,179 @@ protected function db( InputInterface $input, OutputInterface $output ) {
...
+				$output->writeln( "<error>The subcommand $subcommand is not recognized</error>" );
+				return 1;
+			}
+			return 0;
+	}
+
+	/**
+	 * Retrieves path to the working copy of mkcert.
+	 *
+	 * @return string|false Path to the mkcert binary or false if not found.
```
roborourke, line 817:
```suggestion
	 * @return string|null Path to the mkcert binary or null if not found.
```

roborourke (commented):
Nice start, pretty cool that it works like that. There's a few potential gotchas around the platform installation I've commented on and some linter errors to fix up (some dumb ones but hey).

I think next steps are:

- [x] Add support for reading TLDs to generate certs for from config, default to `altis.dev` and maybe `local` and `localhost` for now
- [ ] Update docs, plus FAQ as noted in inline comments
- [x] Run the install and generate subcommands on server start (maybe an option flag or config option to not do this would be useful?)

I'm also thinking in this instance it might be worth punting the code to another class, and making the subcommands into class methods. The same can be said for more of the other commands too so we can maybe refactor at some point. Don't do anything for now, I'm just thinking about code reuse or some of the checks & output.

shadyvb:
@roborourke the domain detection and programmatic generation has been added to #466 , does not factor `localhost` and `local` though, unsure how would those be used, but should be easy enough to add.

inc/composer/class-command.php
```diff
@@ -654,6 +661,179 @@ protected function db( InputInterface $input, OutputInterface $output ) {
...
+						$binary_arch = 'windows-amd64.exe';
+						break;
+					default:
+						$binary_arch = null;
+						break;
+				}
+
+				// If couldn't detect a support architecture, ask the user to install mkcert manually.
+				if ( ! $binary_arch ) {
+					$output->writeln( '<error>This command is only supported on macOS, Linux, and Windows x64, install `mkcert` manually for other systems.</error>' );
```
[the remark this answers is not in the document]
shadyvb: I'd considered we'd let the documentation cover that instead.

inc/composer/class-command.php
shadyvb: Yes, testing cross-platform was supposed to be the next step. I can do that in a VM.

inc/composer/class-command.php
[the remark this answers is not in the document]
roborourke: Docs are often the last thing people might look at, not reason not to be helpful in a timely way and in context.

inc/composer/class-command.php
rmccue: WSL won’t replicate accurately within a VM so far as I know, since it’s a VM itself.

For WSL though, the arch will be Linux because it’s literally a Linux VM you’re running inside of.

roborourke:
@shadyvb yeah sorry - I was just looking there and realised that. Also verified what I was thinking about localhost and local - I was incorrect. I've linked to an avahi-daemon container that we may be able to use alongside the traefik proxy.

My thinking is if we can support .localhost and maybe .local OOTB and document how to add hosts entries for anything else that might be enough.

shadyvb:
We can support ANY TLD when it comes to SSL certificates, if .local / .localhost would alleviate the need for DNS resolution then let's do that, unsure about that though. Either way, outside of the scope here.

inc/composer/class-command.php
shadyvb: If composer is running within WSL, the certificate would be automatically installed to Windows store itself, which is a problem. We'll need to add docs on manually accepting the root CA.

Also, does this mean we do not support Windows itself but rather WSL ? ( I have not used WSL nor Docker on Windows )

roborourke: Yeah there was a PR to support windows natively but I think there were some limitations to that. We do have a way to check it’s a WSL environment at least.

roborourke: You can run .exe files from WSL https://docs.microsoft.com/en-us/windows/wsl/filesystems

inc/composer/class-command.php
shadyvb: The output complains clearly, but we can probably script it as well by checking the output itself.

inc/composer/class-command.php
missjwo: Happy to test stuff - just let me know when the PR is ready to be tested and what i should be looking out for.

inc/composer/class-command.php
shadyvb: Removed in favor of the ending return statement after the select condition.

inc/composer/class-command.php
shadyvb: Seems WSL does have curl by default.

inc/composer/class-command.php
shadyvb: It should do it for all domains, however, we don't have a way to specify _multiple_ domains just yet, but when we do, the command can be configured to do that. Traefik is configured to use that one certificate as the default one.

inc/composer/class-command.php
shadyvb: That's how it works now with the updated PR that introduced the config detection, however, it does not merge with the config, assuming the user knows what they're doing. Can be an improvement though.

shadyvb: And yes you can specify multiple domains, including wildcard domains on the command line.

inc/composer/class-command.php
shadyvb: The command output already includes the location of certificates, so i'll remove the redundancy here.

inc/composer/class-command.php
shadyvb: Moved the call to the beginning, excluding `install` subcommand, to avoid the duplication.

inc/composer/class-command.php
rmccue: @shadyvb Correct, only WSL is supported, per https://docs.altis-dxp.com/local-server/windows/

You can execute exe files, yeah; WSL will transparently map execution of those, and execute them. Note though, `php_uname( 's' )` will return Linux because it is Linux. We need to use the `Command::is_wsl()` function to detect WSL, not uname.

Also happy to test with my WSL env, and screenshare as needed.

inc/composer/class-command.php
roborourke: I think that's fine then. Lets make improvements later if we find things are proving tricky for people.

docker/conf/traefik.toml
```diff
@@ -37,10 +37,10 @@ insecureSkipVerify = true
     address = ":80"
   [entryPoints.https]
     address = ":443"
-    [entryPoints.https.tls]
-      [[entryPoints.https.tls.certificates]]
-      certFile = "/etc/traefik/sni/cert/altis.pem"
-      keyFile = "/etc/traefik/sni/key/altis.pem"
+      [entryPoints.https.tls]
```
roborourke, line 40: Is this indentation change important?

inc/composer/class-command.php
```diff
@@ -211,6 +220,8 @@ protected function get_env() : array {
 	protected function start( InputInterface $input, OutputInterface $output ) {
 		$output->writeln( '<info>Starting...</>' );
 
+		$this->check_ssl_certificate( $input, $output );
```
roborourke, line 223: I don't think we want to do this if the module config has `secure` set to false.

inc/composer/class-command.php
```diff
@@ -654,14 +705,175 @@ protected function db( InputInterface $input, OutputInterface $output ) {
...
+					$output->writeln( '<error>Could not setup `mkcert` properly, try manually installing it.</error>' );
+					$output->writeln( '<error>Download and setup `mkcert` from https://github.com/FiloSottile/mkcert </error>' );
+					return 1;
+				}
+
+				$output->writeln( '<info>mkcert root CA was installed and accepted successfully.</info>' );
+				break;
+			case 'generate':
+				// TODO figure out how to programmatically detect the domains to use.
+				$domains = $input->getArgument( 'options' )[1] ?? '*.altis.dev';
```
roborourke, line 802: I think the default behaviour here should be the following:

- Always generate certs for `<project>.<tld>` and `*.<project>.<tld>` (no harm in them being there by default, even if someone calls the function with other domains, its better to avoid surprises here IMO than assume someone knows it might be a deleterious action if we don't explicitly call that out)
- Add support for a list of hosts in the config and merge those in by default too, we can document that custom ones must be pointed to localhost

inc/composer/class-command.php
```diff
@@ -654,14 +705,175 @@ protected function db( InputInterface $input, OutputInterface $output ) {
...
+
+				exec( "$mkcert -cert-file vendor/ssl-cert.pem -key-file vendor/ssl-key.pem $domains", $dummy, $result );
+
+				if ( $result ) {
+					$output->writeln( '<error>Could not generate certificates! Try generating them manually using `mkcert`.</error>' );
+					$output->writeln( "<error>Command should be: 'mkcert -cert-file vendor/ssl-cert.pem -key-file vendor/ssl-key.pem $domains'</error>" );
+					return 1;
+				}
+
+				$output->writeln( '<info>Generated SSL certificate successfully.</info>' );
```
roborourke, line 812: Will the proxy need to be restarted here? Thinking it might as it's traefik 1.7.

docker/conf/traefik.toml
shadyvb: Not syntactically, no. Was following indentation behavior in previous lines, and probably copy/pasted like that as well.

inc/composer/class-command.php
shadyvb: Not 100% sure I tested that scenario, needs to be tested.

inc/composer/class-command.php
shadyvb: Agreed :+1:

inc/composer/class-command.php
shadyvb: Agreed 👍

inc/composer/class-command.php
shadyvb: Traefik will need to know about the extra domains too 🤔

shadyvb:
@roborourke there is a dilemma here, when the site name is changed in `composer.json`, the naming convention for containers changes, and that brings new containers instead of reusing old ones, and leaves the old ones running. 

Related to the custom domain configuration rather than SSL.

shadyvb:
Related to the above ^, we can instruct people to destroy containers before changing domains, so we don't have to programmatically do that ( we do have a reference to the previous domain, but don't think it's worth the effort ).

shadyvb:
Ah, another issue, since we're generating Traefik's certificate based on the project parameters ( name, tld, extra domains ), the resulting certificate won't really work for other projects, so it'll stop working there IF they're not using the default altis.dev ( which is included in certificate generation now ).

inc/composer/class-command.php
shadyvb: Actually, Traefik will still expect the SSL certificate to exist, so it's safer to just generate the file either way.

inc/composer/class-command.php
shadyvb: Added the extra domains support to nginx ⬆️

shadyvb: Also added the configured domain(s) to the domains within `generate()` command as discussed, also `altis.dev *.altis.dev`.

inc/composer/class-command.php
shadyvb: It needed to be restarted indeed, added logic to restart the proxy service if running, after the generation process.

inc/composer/class-command.php
roborourke: Ah right. Hmm. That's based on the current static traefik.toml file right. I was more conscious of this code running on codespaces, not sure if it'll work or not there.

roborourke:
> Related to the above ^, we can instruct people to destroy containers before changing domains, so we don't have to programmatically do that ( we do have a reference to the previous domain, but don't think it's worth the effort ).

Agreed, that feels like something that won't happen very often, we know from experience teams prefer to configure the local server name and commit that so everyone has a consistent URL structure and can share databases etc... We can maybe add it as an FAQ item.

> Ah, another issue, since we're generating Traefik's certificate based on the project parameters ( name, tld, extra domains ), the resulting certificate won't really work for other projects (other local-server instances running on the same host), so it'll stop working there IF they're not using the default altis.dev ( which is included in certificate generation now ).

That's a good point, hadn't thought about the shared nature of it. Ideally we could generate the certs with project specific names and load them all in from the directory.

While having a `*.altis.dev` cert gets us to where we are now the important new addition is really `*.<project>.altis.dev` (altis.dev on its own is never used locally).

I think we would have to start generating the traefik toml file in PHP, there are parsers out there for this.

Alternatively we would require users to run `composer server start` when switching to another project, and you could not have more than one project working at a time, at least with subdomains.

shadyvb:
Unless.. unless.. we can use container labels to set/aggregate all domains of all setup altis instances... 🤔

roborourke:
> Unless.. unless.. we can use container labels to set/aggregate all domains of all setup altis instances...

Labels on the proxy? Not sure I follow completely, but definitely have a play around!

For configured custom TLDs or custom domains (provided they have been pointed to localhost) we may need to add more traefik frontend rules to the nginx container, or perhaps just extend the existing regex matcher.

I think we can consider any custom TLDs and domains to be experimental at this stage though. I think if we've achieved the primary goals of supporting subsubdomains of altis.dev over https and local server working on codespaces as Ryan noted then we're good.

shadyvb:
> Labels on the proxy? Not sure I follow completely, but definitely have a play around!

Not on the proxy itself, but on containers, i.e. you put the domains as labels on web containers, and then query them while regenerating the certificate via something like `docker ps -q --filter 'label=traefik.port' | xargs -I{} docker inspect {} --format='{{index .Config.Labels "traefik.port"}}'` (substituting `traefik.port` with our custom key e.g. `traefik.domain`).

shadyvb:
.. same can be done if we decide to manage the hosts entry automagically as well, we'd have the list of domains that needs to be added.

shadyvb:
S3, Tachyon, and Analytics services are now working as expected after [81ab150](https://github.com/humanmade/altis-local-server/pull/465/commits/81ab150f970f1e31e1bf85b77f1a3ee893998edf) and [90ce957](https://github.com/humanmade/altis-local-server/pull/465/commits/90ce95706949a38960912af97f0ef5bdc1cf7884)

shadyvb:
Added a notice about missing hosts entries with a full line to copy/paste into /etc/hosts.

Next:
- [ ] Figure out complexity of managing custom domains across all running instances of Altis local-server
- [ ] Document the changes

docker/conf/traefik.toml
shadyvb: Figured out what you meant 🤦 and fixed it.

inc/composer/class-docker-compose-generator.php
```diff
@@ -578,7 +586,8 @@ protected function get_service_tachyon() : array {
 				'environment' => [
 					'AWS_REGION' => 'us-east-1',
 					'AWS_S3_BUCKET' => "s3-{$this->project_name}",
-					'AWS_S3_ENDPOINT' => "https://{$this->tld}/",
+					'AWS_S3_ENDPOINT' => "https://{$this->tld}/s3-{$this->project_name}/",
+					'NODE_TLS_REJECT_UNAUTHORIZED' => 0,
```
roborourke, line 590: TIL, nice find

inc/composer/class-command.php
```diff
@@ -211,7 +220,38 @@ protected function get_env() : array {
...
+			$generated = $this->getApplication()->find( 'local-server' )->run( new ArrayInput( [
+				'subcommand' => 'ssl',
+				'options' => [
+					'generate',
+					'*.altis.dev', // default domain, configured names will be automatically added.
+				],
+			] ), $output );
+
+			if ( $generated ) {
+				exit( 1 );
```
roborourke, line 247:
```suggestion
				return 1;
```

shadyvb:
Further actions:

- [ ] Documentation
- [ ] Add a new issue to support custom domains across different running instances through labels

docs/README.md
```diff
@@ -18,10 +18,51 @@ Navigate your shell to your project's directory. You should already have install
 
 ### Experimental Features
 
+#### Mutagen
+
 You may find that file sharing performance or server response times are slower than you would like on Windows or MacOS. Local Server provides an experimental integration with [Mutagen](https://mutagen.io/) to resolve this.
 
 See the [Mutagen set up guide for detailed instructions on how to install and run it](./mutagen-file-sharing.md).
 
+#### Subdomain and Custom domains in multisites
```
roborourke, line 27: Do we want this under experimental features? Custom domains / codespaces perhaps but I think the SSL stuff itself doesn't need to be considered experimental as we need it to work for the default TLD at the very least.

shadyvb: It's experimental in the sense that it is tied to custom domains which is experimental 🤔

shadyvb: See https://github.com/humanmade/altis-local-server/issues/475

roborourke: I think it’s fine to not list it as experimental personally, codespaces definitely will be to start with

shadyvb: @roborourke Updated docs in https://github.com/humanmade/altis-local-server/pull/465/commits/2f11c98a46527f83c2045563133c2ad948c0a547 , should be ready for a final review.

roborourke (approved):

Issue #277, opened 2021-03-18
Allow "altis.dev" TLD to be configured


PriyaBhatia0210:
In order to support GitHub Codespaces we need to be able to configure the TLD for Local Server. In addition we need to be able to easily provision and manage self signed certificates in addition to supplying the cert for `*.altis.dev`.

Things to note:

- Altis now enforces HSTS by default so we either need to switch it off for codespaces or ensure self signed certificates are setup prior to start Local Server
- Certs have to be mounted on the nginx and traefik containers to function, so need to be setup before Local Server starts

https://github.com/humanmade/altis-local-server/pull/237

PR #466, opened 2022-04-23
Codespaces support, custom domain support, ssl cert support, what else?

shadyvb:
Refresh of https://github.com/humanmade/altis-local-server/pull/341

Rebased on top of SSL changes in https://github.com/humanmade/altis-local-server/pull/465 ( to make a clearer view of changes once that is merged )

Ref https://github.com/humanmade/altis-local-server/issues/277


docker/conf/traefik.toml
```diff
@@ -37,10 +37,10 @@ insecureSkipVerify = true
     address = ":80"
   [entryPoints.https]
     address = ":443"
-    [entryPoints.https.tls]
-      [[entryPoints.https.tls.certificates]]
-      certFile = "/etc/traefik/sni/cert/altis.pem"
-      keyFile = "/etc/traefik/sni/key/altis.pem"
+      [entryPoints.https.tls]
+      [entryPoints.https.tls.defaultCertificate]
+        certFile = "/etc/traefik/ssl-cert.pem"
```
roborourke, line 42: Ah handy! That's useful

inc/composer/class-docker-compose-generator.php
```diff
@@ -131,17 +132,17 @@ protected function get_php_reusable() : array {
 				'ELASTICSEARCH_HOST' => 'elasticsearch',
 				'ELASTICSEARCH_PORT' => 9200,
 				'AWS_XRAY_DAEMON_HOST' => 'xray',
-				'S3_UPLOADS_ENDPOINT' => "https://{$this->tld}/",
+				'S3_UPLOADS_ENDPOINT' => $this->set_url_scheme( "http://s3-{$this->hostname}.localhost/" ),
```
roborourke, line 135: Might have some related comments on the other PR about this, but using just the TLD was required due to the way the SDK generates URLs based on the configured region etc... and the way the Mnio/S3 server handles them to route appropriately.

We probably need to keep it pointing to `$this->tld` rather than localhost. Don't suppose you remember why this was @rmccue ?

inc/composer/class-docker-compose-generator.php
```diff
@@ -557,7 +558,7 @@ protected function get_service_tachyon() : array {
 				'environment' => [
 					'AWS_REGION' => 'us-east-1',
 					'AWS_S3_BUCKET' => "s3-{$this->project_name}",
-					'AWS_S3_ENDPOINT' => "https://{$this->tld}/",
+					'AWS_S3_ENDPOINT' => $this->set_url_scheme( "http://{$this->tld}/" ),
```
roborourke, line 561: Ah maybe this was the problem re. the S3 endpoint environment variable in the PHP container config. It might just need to match the new value.

inc/namespace.php
```diff
@@ -87,6 +91,15 @@ function bootstrap() {
 	// Filter ES package IDs for local.
 	add_filter( 'altis.search.packages_dir', __NAMESPACE__ . '\\set_search_packages_dir' );
 	add_filter( 'altis.search.create_package_id', __NAMESPACE__ . '\\set_search_package_id', 10, 3 );
+
+	// If we're on Codespaces, the native host will be localhost.
+	if ( $config['codespaces_integration'] ?? null && $_SERVER['HTTP_HOST'] === 'localhost' ) {
```
roborourke, line 96:
```suggestion
	if ( ( $config['codespaces_integration'] ?? null ) && $_SERVER['HTTP_HOST'] === 'localhost' ) {
```

roborourke (commented):
Couple of notes inline.

DNS resolution is a pain, I was hoping as traefik runs on localhost:80 that might just work. Perhaps the Chassis approach with avahi might work https://github.com/damonmorgan/docker-mdns

(at least with docker we can ensure it's running!)

inc/composer/class-docker-compose-generator.php
roborourke: See this comment before making any changes https://github.com/humanmade/altis-local-server/pull/466#discussion_r857812295

inc/namespace.php
```diff
@@ -28,13 +28,17 @@ function bootstrap() {
 		add_filter( 's3_uploads_s3_client_params', function ( $params ) {
 			if ( defined( 'S3_UPLOADS_ENDPOINT' ) && S3_UPLOADS_ENDPOINT ) {
 				$params['endpoint'] = S3_UPLOADS_ENDPOINT;
+				$params['bucket_endpoint'] = true;
```
rmccue, line 31: @roborourke This fixes the S3 SDK for the issue you mentioned.

inc/composer/class-docker-compose-generator.php
rmccue: Yes; I fixed this separately, see line I commented on in inc/namespace.php

shadyvb:
Closing down in favor of separating concerns between SSL and Codespaces.

PR #473, opened 2022-04-28
Update Local Server to use PHP 8 (alpha container, no v8js)

### LINKED ISSUE #437: Update local-server to PHP 8
Dependent on the availability of the updated docker image.

Acceptance criteria:

- [x] Update default PHP image to version with PHP 8
- [x] Allow configuring to use PHP 7.4 instead
- [x] Configuration options are documented

roborourke:
Fixes #437

Note - removes v8js support.

### DIFF
--- a/docs/php-version.md
+++ b/docs/php-version.md
@@ -1,0 +1,21 @@
+# PHP Version
+
+In some cases an Altis version may provide support for multiple versions of PHP. This can be useful for testing and developing while updating an application to work on the latest available version of PHP.
+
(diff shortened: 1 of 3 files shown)


shadyvb:
@roborourke Do we want to make 8.0 the default in trunk ? shouldn't we make it opt-in rather than opt-out until cleared stable ?

roborourke:
v12 needs to run PHP 8 as stable, which is effectively what `trunk` is. We may backport this option to v11, and if we do then it would default to 7.4 and people could opt in.

shadyvb (approved):

Issue #477, opened 2022-05-09
Requests to self fail because of self-signed SSL certificate


shadyvb:
PHP does not know about the custom SSL certificate local-server generates, so it fails while requesting any self URL.

Steps to reproduce:
1. `wp_remote_get( $main_domain )`

Acceptance criteria:

- [x] Requesting self URLs should skip validating the SSL certificate.

PR #478, opened 2022-05-09
Skip validating SSL on local requests

shadyvb:
Ref https://github.com/humanmade/altis-local-server/issues/477


roborourke (approved):
Do we need to worry about subdomains too? Didn’t realise this would be a problem if SSL is terminated at traefik!

shadyvb:
We could add the CA cert to the container, but that has to happen in runtime rather than build time, and with the slim image we have, that's not straight-forward. Should we attempt that instead ?

And yeah, good point about the subdomains, I had intended to support them but forgot, will update.

PR #479, opened 2022-05-12
Automatically install mkcert if not yet available

roborourke:
Should allow travis runs to continue as before. I noticed because the module.yml in dev tools was missing the `composer server ssl install` step that all the tests started failing.

This might resolve the output Mike saw here https://github.com/humanmade/product-dev/issues/1039


shadyvb (approved):

roborourke:
You're handling what happens if the new `secure` config option is set to false in the codespaces work iirc? Just made me think of it working on this.

PR #480, opened 2022-05-12
Ensure server start works with an existing install of mkcert

roborourke:
Changes the return code to zero to allow execution to proceed. `mkcert` already being available doesnt need to count as a failure for `composer server sll install`.


robindevitt (approved):
Looks good 👍🏼

PR #486, opened 2022-06-10
Switch to latest tag for mailhog

roborourke:
The currently pinned tag has been removed from docker hub so builds are failing

### DIFF
--- a/inc/composer/class-docker-compose-generator.php
+++ b/inc/composer/class-docker-compose-generator.php
@@ -604,7 +604,7 @@ class Docker_Compose_Generator {
 	protected function get_service_mailhog() : array {
 		return [
 			'mailhog' => [
-				'image' => 'cd2team/mailhog:1632011321',
+				'image' => 'cd2team/mailhog:latest',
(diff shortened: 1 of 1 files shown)


wisyhambolu (approved):

Sephsekla:
Travis is failing on this with

```
Error: Error establishing a database connection.

The command "cd $HOME/test-root && composer server cli site list" exited with 1.
```

tfrommen (approved):
I don't know if the previous reference was an exact tag for a reason or not, but if we're OK with pulling in the latest, this is looking good.

roborourke:
Yeah mailhog is a fairly low blast radius if something does happen - `latest` is literally the only tag and I'm not sure pinning to a specific sha commit value is 100% reliable right now either.

goldenapples (approved):
Thanks for fixing this!

roborourke:
@Sephsekla local server intermittently fails to come up on Travis on the first run sometimes, I think it's a memory issue and I'm hoping switching to use Arm based travis builds will help.

Issue #497, opened 2022-06-23
Error when using without Altis


mattheu:
We're using Altis local server as a standalone local development environment for a non-altis project. This is working great with v11.0.1 but is broken in v11.0.4. 

Whilst I understand this isn't officially 'supported' it is documented in the readme of this project as a way it can be used, so I'd not expect to have a hard dependency on Altis core. 

Steps to reproduce:

Install local server as a standalone environment. [As documented in the readme](https://github.com/humanmade/altis-local-server#installation-without-altis). Run `composer server start. `

I then get the following error: 

```Call to undefined function Altis\get_config() ```

Note this worked fine on v11 but there has been a breaking change in a security/patch update.

**Comment by Sephsekla** [comment]
Just ran into this issue myself while trying out using local server to test WordPress 6.0.

The problematic line is https://github.com/humanmade/altis-local-server/blob/master/inc/composer/class-docker-compose-generator.php#L843, I was able to workaround locally by changing the following:

```php
		$modules = Altis\get_config()['modules'] ?? [];
```

⬇️ 

```php
		$modules = [];

		if ( function_exists( 'Altis\\get_config' ) ) {
			$modules = Altis\get_config()['modules'] ?? [];
		}
```

Happy to open a PR if the Altis team are OK with this as a fix.

mattheu:
Whilst this stops the blocking error, the problem with this suggestion @Sephsekla is that it doesn't give you the exact same functionality that you had before this breaking change was introduced. 

Looking at https://github.com/humanmade/altis-local-server/commit/21639f5314148fedf8c5a6e2cc193c538e0bb468, it used to read from composer.json. So probably we need something like

```
		if ( function_exists( 'Altis\\get_config' ) ) {
			$modules = Altis\get_config()['modules'] ?? [];
		} else {
			$json = file_get_contents( $this->root_dir . DIRECTORY_SEPARATOR . 'composer.json' );
			$composer_json = json_decode( $json, true );
			$modules = ( $composer_json['extra']['altis']['modules'] ?? [] );
		}
```

Sephsekla:
Oh that's a good point - I was just using it with WP Core so hadn't considered that.

What's the use case for loading Altis modules but not Altis itself? At that point are you not essentially using Altis with some custom configuration applied?

robindevitt:
I came across the same issue and did the following steps to get me going:
- `composer update -W`
- `composer self-update` and then `composer update`
- `composer server start`

[...]

PR #502, opened 2022-07-16
Manually load altis/core in Composer plugin

### LINKED ISSUE #501: Call to undefined function Altis\get_config()
One of the issues that's been popping off lately is this uncaught fatal error in Local Server:

```
Fatal error: Uncaught Error: Call to undefined function Altis\get_config() in /path/to/local-server/inc/composer/class-docker-compose-generator.php:843
Stack trace:
#0 /path/to/local-server/inc/composer/class-docker-compose-generator.php(89): Altis\Local_Server\Composer\Docker_Compose_Generator->get_config()
#1 /path/to/local-server/inc/composer/class-docker-compose-generator.php(205): Altis\Local_Server\Composer\Docker_Compose_Generator->get_php_reusable()
#2 /path/to/local-server/inc/composer/class-docker-compose-generator.php(718): Altis\Local_Server\Composer\Docker_Compose_Generator->get_service_php()
#3 /path/to/local-server/inc/composer/class-docker-compose-generator.php(829): Altis\Local_Server\Composer\Docker_Compose_Generator->get_array()
#4 /path/to/local-server/inc/composer/class-command.php(9 in /path/to/local-server/inc/composer/class-docker-compose-generator.php on line 843
```

I've been digging a bit more into this after seeing unit tests broken with Composer 1.x. This is the offending commit: 16260c7d65dad7e824b36db39f2e9d5244fdddaa where we went from reading the JSON file directly, to using a function defined in Altis Core. There's also #497 which was caused by the same commit, and could likely be solved by using the latest version of Composer and `altis/core`.

However, there is still a problem with Composer 1.x, as well as anything earlier than Composer 2.2.0, where the `files` section [used in the](https://github.com/humanmade/altis-core/blob/89d59f26918c017218b9def6af1a8b4e622d1123/composer.json#L20) `altis/core` autoload setting are not being loaded at all when running Composer plugins. This is a bug that seems to be fixed in Composer 2.2.0, here's the related PR: https://github.com/composer/composer/pull/10065

I was able to verify this by running `composer self-update 2.1.14` and still seeing the error upon `composer server start`.

Unless we're ready to drop support for Composer < 2.2.x we should look into fixing this. Classes autoloading still works as expected, so moving the `get_config()` logic into an `Altis\Config` class in altis/core may be an option. Another option would be to continue using `file_get_contents()`.

Acceptance criteria:

- [ ] `composer server start` works with Composer 1.x and 2.1.x

rmccue:
Fixes #501.

Using Composer 2.1.9 locally, confirmed this fixes the command for me.

### DIFF
--- a/inc/composer/class-plugin.php
+++ b/inc/composer/class-plugin.php
@@ -26,6 +26,17 @@ class Plugin implements PluginInterface, Capable {
 	 * @return void
 	 */
 	public function activate( Composer $composer, IOInterface $io ) {
+		if ( ! function_exists( '\\Altis\\get_config' ) ) {
+			// Composer <2.2 has broken autoloading, so we need to manually
+			// load altis/core in.
+			$vendor_altis = dirname( __DIR__, 3 );
+			$path = $vendor_altis . '/core/inc/namespace.php';
+			if ( file_exists( $path ) ) {
+				require $path;
+			} else {
+				trigger_error( 'Altis\\get_config() not found, and cannot manually load altis/core module. See https://github.com/humanmade/altis-local-server/issues/501', E_USER_WARNING );
+			}
+		}
 	}
 
 	/**
--- a/inc/composer/class-plugin.php
+++ b/inc/composer/class-plugin.php
@@ -29,8 +29,7 @@ class Plugin implements PluginInterface, Capable {
 		if ( ! function_exists( '\\Altis\\get_config' ) ) {
 			// Composer <2.2 has broken autoloading, so we need to manually
 			// load altis/core in.
-			$vendor_altis = dirname( __DIR__, 3 );
-			$path = $vendor_altis . '/core/inc/namespace.php';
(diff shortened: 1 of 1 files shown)


roborourke (approved):
Should we add it as a peer dep suggestion or requirement with version * too?

kovshenin (approved):

kovshenin:
> Should we add it as a peer dep suggestion or requirement with version * too?

If this package is not intended to be used on its own, I don't think we need to add the requirement here, or to any other Altis module that uses `Altis\get_config()`. We should also remove the "Installation without Altis" section from the readme if that's the case, then for #497 we can suggest trying to `composer require --dev altis/core` with no guarantee that it will continue working in the future.

rmccue:
If it's not onerous for us, we can add the dep, but otherwise I'd say whatever is least painful for us.

kovshenin (approved):

Issue #508, opened 2022-08-01
Can't sendmail through WP-CLI on Linux


kovshenin:
On Linux, the `composer server cli` command runs in a PHP container using the host system's user, typically 1000. This has been part of the work in #183 to support filesystem access on volumes mounted directly into the containers, retaining the uid/gid. However, with this change some programs may fail to run, because the user 1000 does not exist in the target container. Sendmail is one of those programs:

```
$ composer server cli -- shell
wp> wp_mail( 'foo@bar.org', 'foo', 'bar' );
sendmail: unknown uid 1000
=> bool(false)
```

Works in `composer server ssh` because it runs through `->shell()` and thus the default container user (www-data), and not `->exec()` where we set the user explicitly.

```
$ composer server ssh
$ id
uid=33(www-data) gid=33(www-data) groups=33(www-data)
$ php -a
php > var_dump( mail( 'foo@bar.org', 'foo', 'bar' ) );
bool(true)
```

For reference in exec (used for running WP-CLI commands as well):

```
$ composer server exec -- id
uid=1000 gid=0(root) groups=0(root)
```

Acceptance criteria:

- [ ] Sendmail succeeds in a WP-CLI command (invoked through `mail()` or `wp_mail()`)
- [ ] The message is delivered to the Mailhog instance

kovshenin:
Fixed in #519 by removing the use of sendmail altogether. Messages are now sent to MailHog directly from `wp_mail -> PHPMailer`, so invoking `wp_mail()` in the CLI container now works as expected.

PR #519, opened 2022-08-09
Route e-mails to MailHog via PHPMailer

### LINKED ISSUE #516: Bug with sendmail substitute
I'm not sure which `sendmail` substitute we used, but it looks like it has a bug.

When Workflows sends an email that is caught by Mailhog, Mailhog fails to parse the sender and subject:

Looking at the raw email headers there is a rogue new line which splits the transport headers from those added by PHP.

This may be a bug in the `sendmail` clone or a misconfiguration. In particular, I have seen this type of erroer where a configuration file specifying headers had DOS line endings in a Unix environment.

kovshenin:
Instead of relying on sendmail to do the work, set PHPMailer to speak SMTP directly to the MailHog service. This removes the sendmail dependency, and resolves line-endings inconsistencies between PHP 7.4 and 8.0.

Sending e-mails directly through `mail()` will no longer work. All sending should be done through `wp_mail()` instead.

Fixes #516

### DIFF
--- a/inc/composer/class-docker-compose-generator.php
+++ b/inc/composer/class-docker-compose-generator.php
@@ -172,7 +172,7 @@ class Docker_Compose_Generator {
 				'S3_UPLOADS_REGION' => 'us-east-1',
 				'S3_CONSOLE_URL' => Command::set_url_scheme( "https://s3-console-{$this->hostname}" ),
 				'TACHYON_URL' => Command::set_url_scheme( "{$this->url}tachyon" ),
-				'PHP_SENDMAIL_PATH' => '/usr/sbin/sendmail -t -i -S mailhog:1025',
+				'PHP_SENDMAIL_PATH' => '/bin/false',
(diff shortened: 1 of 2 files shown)


rmccue:
> Sending e-mails directly through `mail()` will no longer work. All sending should be done through `wp_mail()` instead.

Will this cause compatibility issues with plugins? 🤔 I guess our production servers don't allow the use of `mail()` anyway so probably closer to the "real" thing in that regard.

kovshenin:
Yeah, this is slightly closer to production. I don't anticipate any compatibility issues as WordPress plugins and themes should never be using `mail()` directly anyway. I think this also fixes #508 since it will no longer be dependent on the current user.

rmccue:
Biggest downside I guess is that we can no longer send _real_ emails when we need to.

rmccue:
> Also not quite sure why you'd want to send actual mail from a dev environment, which is pretty much guaranteed to end up in spam or rejected anyway.

I don't judge 😛 It is legitimately useful if you need to test email rendering eg in Gmail.

That said, YAGNI.

shadyvb (approved):

roborourke (approved):
Nice, wasn't aware of this option

PR #540, opened 2022-11-04
Add a command for cloud-build

joehoyle:
So people can more easily test the builds as if there were in cloud, I've added a command that will run their `.build-script` using the same `build-container` that we used in CodeBuild.

### DIFF
--- a/inc/composer/class-command.php
+++ b/inc/composer/class-command.php
@@ -172,6 +172,8 @@ EOT
 			return $this->shell( $input, $output );
 		} elseif ( $subcommand === 'import-uploads' ) {
 			return $this->import_uploads( $input, $output );
+		} elseif ( $subcommand === 'cloud-build' ) {
+			return $this->cloud_build( $input, $output );
 		} elseif ( $subcommand === null ) {
 			// Default to start command.
 			return $this->start( $input, $output );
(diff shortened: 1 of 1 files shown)


inc/composer/class-command.php
```diff
@@ -733,6 +735,35 @@ protected function import_uploads() {
...
 
+	/**
+	 * Run the .build-script as it would be in Altis Cloud
+	 *
+	 * @param InputInterface $input
+	 * @param OutputInterface $output
+	 * @return void
+	 */
+	protected function cloud_build( InputInterface $input, OutputInterface $output ) {
+		$output->writeln( '<info>Buld will install production dependencies, so you may need to `composer install` when doing development again.</info>' );
```
rmccue, line 746:
```suggestion
		$output->writeln( '<info>Build will install production dependencies, so you may need to `composer install` when doing development again.</info>' );
```

rmccue (commented):
Is it possible to have this build into a temp-/sub-directory instead of overwriting the current dir?

inc/composer/class-command.php
```diff
@@ -172,6 +172,8 @@ protected function execute( InputInterface $input, OutputInterface $output ) {
 			return $this->shell( $input, $output );
 		} elseif ( $subcommand === 'import-uploads' ) {
 			return $this->import_uploads( $input, $output );
+		} elseif ( $subcommand === 'cloud-build' ) {
```
rmccue, line 175: Suggestion: `check-build`?

kovshenin:
> Is it possible to have this build into a temp-/sub-directory instead of overwriting the current dir?

I was going to ask the same and went down a rabbit hole into how to force a volume mount with an overlay filesystem. Turned out to be a bit trickier than I though, but copying into a non-host directory and running the build there would also be okay I suppose.

rmccue:
> Turned out to be a bit trickier than I though, but copying into a non-host directory and running the build there would also be okay I suppose.

Yeah, I'd copy it probably.

joehoyle:
Hmm yeah if we want this to be a "readonly" operation, we'd need to do things a bit differently. We _could_ increase the scope to run the "build process" from Altis Cloud (not just the build script), so that would mean: `git clone`, etc. In a subdir that's cleaned up afterwards or something. We _could_ copy the entire working directory, but we'd want to exclude things like uploads, and existing node_modules _could_ present another area of "not like for like" with production.

kovshenin:
That's one of the reasons I was hoping to have an overlay filesystem. With `git clone` it would mean that any changes have to be committed and pushed before testing, and the biggest advantage of having this available in Local Server is being able to test builds before committing your work.

Doing a copy and excluding things in .gitignore may be a good balance.

inc/composer/class-command.php
tfrommen: Why not just `build`?

kovshenin: I think the intention is to test that a build runs without errors, and doesn't actually produce a build that can be deployed to Altis, so `build` may be a bit misleading.

rmccue:
What if we make it the user's choice? By default, `cp -r` the working directory; allow a `--commit[=HEAD]` option.

kovshenin:
Wouldn't the commit still have to exist in the remote? And if so, wouldn't it likely already have a remote build available/running in Vantage?

rmccue:
Not necessarily, it can use your local commits which may not have been pushed; additionally, you're likely pushing to a feature branch which isn't getting built in Vantage, not one of the deploy branches.

PR #561, opened 2023-01-12
Set redis container image version to v7.0

wisyhambolu:
The Redis container is currently set to use a rather old version of Redis (v3.2), this sets the Redis container to use the latest Redis version (v7.0) which brings multiple security fixes and speed improvements.


mikelittle (approved):
Looks good to me

PR #574, opened 2023-02-03
Avoid warnings for mkcert

### LINKED ISSUE #573: Warnings/error output when mkcert isn't installed on the system
Steps to reproduce:
1. Ensure mkcert is not installed system-wide
2. Run any command which invokes mkcert, such as `composer server ssl`

A pristine output.

Instead:
```
$ composer server ssl
sh: mkcert: command not found
Deprecation Notice: trim(): Passing null to parameter #1 ($string) of type string is deprecated in .../vendor/altis/local-server/inc/composer/class-command.php:900
sh: vendor/mkcert: No such file or directory
Deprecation Notice: trim(): Passing null to parameter #1 ($string) of type string is deprecated in .../vendor/altis/local-server/inc/composer/class-command.php:906
sh: mkcert: command not found
Deprecation Notice: trim(): Passing null to parameter #1 ($string) of type string is deprecated in .../vendor/altis/local-server/inc/composer/class-command.php:900
sh: vendor/mkcert: No such file or directory
Deprecation Notice: trim(): Passing null to parameter #1 ($string) of type string is deprecated in .../vendor/altis/local-server/inc/composer/class-command.php:906
sh: mkcert: command not found
sh: vendor/mkcert: No such file or directory
```

Acceptance criteria:

- [ ] Running commands without mkcert installed should not trigger PHP warnings
- [ ] stderr output from sh should not be displayed to user for detection commands

rmccue:
Fixes #573

### DIFF
--- a/inc/composer/class-command.php
+++ b/inc/composer/class-command.php
@@ -898,13 +898,13 @@ EOT;
 
 		// Check if mkcert is installed globally already, bail if so.
 		$version = trim( shell_exec( 'mkcert -version 2>/dev/null' ) ?? '' );
-		if ( $version ) {
+		if ( strlen( $version ) > 0 ) {
 			return 'mkcert';
 		}
 
 		// Check if mkcert is installed locally already, bail if so.
 		$version = trim( shell_exec( "$mkcert -version 2>/dev/null" ) ?? '' );
-		if ( $version ) {
+		if ( strlen( $version ) > 0 ) {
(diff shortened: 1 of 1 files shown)


tomjn (approved):
I'd have used `empty` instead of `strlen` personally but this works

mikelittle (approved):

rmccue:
Worth backporting?

PR #593, opened 2023-03-17
Support Afterburner

joehoyle:
We need to define the constants used by https://github.com/humanmade/altis-cloud/pull/731

See https://github.com/humanmade/product/issues/567


mikelittle (approved):
Looks good to me.

kovshenin (approved):

Issue #592, opened 2023-03-15
Add support for PHP8.2


mikelittle:
## Acceptance criteria:
- [ ] Local server can run PHP 8.2
- [ ] Works as expected with PHP 8.2 on the host system

kovshenin:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Vulcan-6284e0dfb386b00c83981426/issues/humanmade/altis-local-server/592?planning-poker) with Zenhub @ferschubert-hm @jerico @mikelittle @wisyhambolu

PR #594, opened 2023-03-31
Add support for PHP8.2

mikelittle:
This addresses issue https://github.com/humanmade/altis-local-server/issues/592


kovshenin (approved):
Looks good. We can add stable release tags for 8.2 and 8.1 during release so they no longer point to beta tags.

PR #608, opened 2023-07-25
Clarify the bucket name in the docker-compose file.

mikelittle:
Make it clearer how the bucket name is derived and used.


roborourke (approved):
Useful addition :)

PR #618, opened 2023-08-16
Throw an exception when asking for db information but the container isn't running

johnbillion:
If you run `composer server db info` and the container isn't running, you'll get a PHP warning about an undefined array key:

This change introduces an exception with a more helpful error message that's thrown if the database container information can't be fetched:


mikelittle (approved):
Good catch. Thanks John.

jerico (approved):

PR #626, opened 2023-08-25
Add option enable afterburner

wisyhambolu:


rmccue:
Can we default this on?

joehoyle:
Sorry I didnt see this, and also created #634 

i think it's better to have it as a composer.json value, as really if it's enabled in Cloud the whole team should have it switched on in version control.

joehoyle:
Closed in favor of #634

PR #634, opened 2023-10-25
Add Afterburner support to Local Server

joehoyle:


docs/using-afterburner.md
```diff
@@ -0,0 +1,19 @@
+# Using Afterburner
+
+If your project uses Afterburner in Altis Cloud, you can enabled it in Altis Local Service by setting the `altis.modules.local-server.afterburner` value to `true`:
```
mikelittle, line 3: "enable" not "enabled"

mikelittle (changes requested):
Looks good. Just a typo in the docs. And a lint issue about quotes.

docs/using-afterburner.md
joehoyle: Ah good catch! Fixed now

mikelittle (approved):
There's still a linter issue about double quotes vs single quotes, but otherwise, looks good to me

jerico (approved):

PR #652, opened 2023-11-24
Add ability to run shell as root

### LINKED ISSUE #112: Allowing opening shell as root / sudo user
To allow people to debug / install packages, they need a root permissions level user in the `shell` command. Passing `-u root` to the command would suffice.

Acceptance criteria:

- [x] Add an option flag to the shell command called `--root`, optional, alias as `-r`
- [x] When flag is set add `-u root` to the `docker exec` call
- [x] Update the commands documentation with this argument
- [x] Add a note to the command docs on why you would use the `root` user:
  - To test modifications to the PHP container eg. installing PHP libraries or other packages
  - To carry out any other tasks that require `sudo` privileges such as making directories outside of the app directory

rmccue:
This is necessary for installing packages/etc.

We should _also_ add sudo to our container.

Fixes #112.


joehoyle (approved):

rmccue (commented):
Noting I've reviewed and approved the updates here, although in the future let's make sure changes are re-reviewed after substantial changes.

PR #682, opened 2024-05-01
Update example to v8

mikelittle:
Updates the info example to show MySQL v8 and replace mention of Sequel Pro with Sequel Ace.

### DIFF
--- a/docs/database.md
+++ b/docs/database.md
@@ -39,13 +39,13 @@ User:           wordpress
 Password:       wordpress
 
 Host:           0.0.0.0
-Port:           32775
+Port:           32809
 
-Version:        5.7.26-1debian9
-MySQL link:     mysql://wordpress:wordpress@0.0.0.0:32775/wordpress
+Version:        8.0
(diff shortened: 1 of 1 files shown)


rmccue:
> replace mention of Sequel Pro with Sequel Ace

Does Sequel Ace actually work with this? I recall trying TablePlus but it had some issues with macOS sandboxing with our integration.

mikelittle:
Yeah. Sequel ace picks up the old sequel pro url style. I use it locally. And I tested it from Local server yesterday,.

mikelittle:
I should add support for table plus at some point because ddev does that so I can see howe they do it.

rmccue:
The issue at least with TablePlus was that macOS' sandboxing wouldn't allow access to the .spf file that we generate as it was outside of the app's sandboxed directory; I think this will apply to any of the apps installed via the Mac App Store?

(FWIW, I still use Sequel Pro.)

mikelittle:
DDEV solved it like this (Mac only)
https://github.com/ddev/ddev/pull/2526/files

```shell
open mysql-dsn -a /path-to-tableplus-app
```

Issue #265, opened 2021-02-09
New sites and subdomains with hyphens throw an Nginx 404 error


jazzsequence:
As part of the testing for https://github.com/humanmade/altis-cms/pull/333, some of the URLs I was trying to test would throw an Nginx (not WordPress) 404 error. The common thread amongst those were that they all had hyphens (`-`) in the domain or subdomain. Some example domains:

* `hyphenated-subdomain.altis-dev.altis.dev`
* `h-y-p-h-e-n-a-t-e-d-s-u-b-d-o-m-a-i-n.altis-dev.altis.dev`
* `two--hyphens.altis-dev.altis.dev`
* `xn--rb-fka.com`
* `xn--rb-fka.altis-dev.altis.dev`
* `xn--jp-cd2fp15c.jp`
* `xn--99zt52a.altis-dev.altis.dev`

Steps to reproduce:
1. Navigate to the Add New Site page in the Altis Network Admin
2. Attempt to create a new site with hyphens either in the subdomain (e.g. `some-site.altis-dev.altis.dev`) or in the custom domain (e.g. `some-site.com`).
3. Navigate to the site on the front end.

Note that for testing, the Google Chrome easter egg (`thisisunsafe`) has to be used to get past the SSL certificate security warning.

Acceptance criteria:

- [x] Creating a new subdomain with a hyphen allows navigating to the front of the site
- [x] Creating a new custom domain with a hyphen allows navigating to the front of the site

PR #692, opened 2024-05-28
Fixing 404s on subdomains with dashes on it

filter182:
Fixed the regex to allow dashes on subdomains, previous regex didn't march them. This was reported on issue [#265](https://github.com/humanmade/altis-local-server/issues/265)

[ML] To validate I created a range of subdomains with hyphens.

CleanShot 2024-12-17 at 16 12 50@2x

---

### Completion Checklist

Is this ticket done? See
[the Play Book Definition of Done](https://playbook.hmn.md/play/product/definition-of-done-2/)

- [x] Has the acceptance criteria been met?
- [x] Is the documentation updated (including README)?
- [x] Do any code/documentation changes meet project standards?
  - [x] Or are manual tests documented (at least on this ticket)?
- [x] Are any Playbook/Handbook pages updated?
- [x] Has a new module release (patch/minor) been created/scheduled?
- [x] Have the appropriate `backport` labels been added to the PR?
- [ ] Is there a roll-out (and roll-back) plan if required?

### DIFF
--- a/inc/composer/class-docker-compose-generator.php
+++ b/inc/composer/class-docker-compose-generator.php
@@ -372,7 +372,7 @@ class Docker_Compose_Generator {
 					'traefik.port=8080',
 					'traefik.protocol=https',
 					'traefik.docker.network=proxy',
-					"traefik.frontend.rule=HostRegexp:{$this->hostname},{subdomain:[A-Za-z0-9-_]+}.{$this->hostname}{$domains}",
+					"traefik.frontend.rule=HostRegexp:{$this->hostname},{subdomain:[A-Za-z0-9.-]+}.{$this->hostname}{$domains}",
 					"traefik.domain={$this->hostname},*.{$this->hostname}{$domains}",
 				],
 				'environment' => [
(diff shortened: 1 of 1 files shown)


inc/composer/class-docker-compose-generator.php
```diff
@@ -327,7 +327,7 @@ protected function get_service_nginx() : array {
 					'traefik.port=8080',
 					'traefik.protocol=https',
 					'traefik.docker.network=proxy',
-					"traefik.frontend.rule=HostRegexp:{$this->hostname},{subdomain:[a-z.-_]+}.{$this->hostname}{$domains}",
+					"traefik.frontend.rule=HostRegexp:{$this->hostname},{subdomain:[A-Za-z0-9-_]+}.{$this->hostname}{$domains}",
```
kovshenin, line 330: I think this fix has some unintended consequences.

The original regex `[a-z.-_]` is quite confusing, but it is two groups: `a through z` and `. through _`. If you look at the ASCII table between `.` and `_` (positions 46-95) you'll see all numbers, uppercase latin letters and a bunch of symbols as well, including the `.` and `_`, but excluding the `-` symbol, which is at position 45.

I'm not sure why the original regex is so permissive in the first place:

```
php > var_dump( preg_match( '#^[a-z.-_]+$#', 'what/a.weird_sub^domain' ) );
int(1)
```

And yet fails to match:

```
php > var_dump( preg_match( '#^[a-z.-_]+$#', 'perfectly-legit' ) );
int(0)
```

While much more restrictive, the proposed newer version fails to match the dot symbol:

```
php > var_dump( preg_match( '#^[A-Za-z0-9-_]+$#', 'foo.bar' ) );
int(0)
```

This might be useful for deeper-level subdomains (language subdomains on a Multisite install for example). It also still matches the underscore symbol, which is not valid for hostnames/subdomains. It also looks like case sensitivity is handled by the HostRegexp implementation itself https://github.com/traefik/traefik/issues/3930

kovshenin (changes requested):

inc/composer/class-docker-compose-generator.php
rmccue: I think the original regex is accidental, and it *should* have been `[a-z.\-_]+`, but the escaping was missed.

filter182: Good catch with the case sensitive implementation and  I totally forgot to include the dot, this one `[a-z.\-]+` works but it doesn't match numbers, do you guys think I should use `[a-z0-9.\-]` (removing the underscore) instead? It's common to see some people that add numbers on testing sites like `test-site-1.domain.com`, etc.

filter182:
> All good now. Thanks for this @filter182

Thanks for fixing it @mikelittle

PR #710, opened 2024-07-09
Add Node.js Support

jerico:
To enable Node.js in Local Server, add the following configuration in the project's `composer.json`:
```json
{
  "extra": {
    "altis": {
      "modules": {
        "local-server": {
          "nodejs": {
            "path": "../altis-nodejs-skeleton"
          }
        }
      }
    }
  }
}
```

`path` refers to the frontend code directory

- [x] Add Node.js container when Node.js configuration exists
- [x] Add documentation
- [x] ~~Add healthcheck~~
- [x] Echo Node.js URL when startup completes
- [x] Add config to setup Node.js version to use
     - Read from ~~`extra.altis.modules.local-server.nodejs.version`~~ `{$app}/package.json`
     - Based off how we set version in TFA https://github.com/humanmade/terraform-app-stack/blob/5a70129f5699b663347be2c2919b63ba98ab6015/modules/nodejs/buildspec.yml#L18

https://github.com/humanmade/product-dev/issues/1589

### DIFF
--- a/docs/nodejs.md
+++ b/docs/nodejs.md
@@ -8,18 +8,18 @@ Node.js can be enabled in Local Server by adding `extra.altis.modules.local-serv
 
 ```json
 {
-	"extra": {
-		"altis": {
-			"modules": {
-				"local-server": {
-					"nodejs": {
-            "path": "../altis-nodejs-skeleton",
-            "version": "21.1"
-          }
-				}
-			}
-		}
-	}
+   "extra":{
+      "altis":{
+         "modules":{
+            "local-server":{
+               "nodejs":{
+                  "path":"../altis-nodejs-skeleton",
+                  "version":"21.1"
+               }
+            }
+         }
+      }
+   }
 }
 ```
 
--- a/docs/nodejs.md
+++ b/docs/nodejs.md
@@ -1,0 +1,27 @@
+# Node.js
+
+Altis supports running Node.js applications alongside WordPress, utilizing WordPress as a headless API.
+
+## Enabling Node.js in Local Server
+
+Node.js can be enabled in Local Server by adding `extra.altis.modules.local-server.nodejs` in the project's `composer.json`
+
+```json
+{
+	"extra": {
+		"altis": {
+			"modules": {
+				"local-server": {
+					"nodejs": {
+            "path": "../altis-nodejs-skeleton"
+          }
+				}
+			}
+		}
+	}
+}
+```
+
+path refers to the relative path of the project's front-end code.
+
+This will make the application available at nodejs-my-project.altis.dev.
--- a/docs/nodejs.md
+++ b/docs/nodejs.md
@@ -13,7 +13,8 @@ Node.js can be enabled in Local Server by adding `extra.altis.modules.local-serv
 			"modules": {
 				"local-server": {
 					"nodejs": {
-            "path": "../altis-nodejs-skeleton"
+            "path": "../altis-nodejs-skeleton",
+            "version": "21.1"
           }
 				}
 			}
@@ -22,6 +23,7 @@ Node.js can be enabled in Local Server by adding `extra.altis.modules.local-serv
 }
 ```
 
-path refers to the relative path of the project's front-end code.
+`path` refers to the relative path of the project's front-end code.
+`version` refers to the version of Node.js to use
 
 This will make the application available at nodejs-my-project.altis.dev.
--- a/docs/nodejs.md
+++ b/docs/nodejs.md
@@ -13,8 +13,7 @@ Node.js can be enabled in Local Server by adding `extra.altis.modules.local-serv
          "modules":{
             "local-server":{
                "nodejs":{
-                  "path":"../altis-nodejs-skeleton",
-                  "version":"21.1"
+                  "path":"../altis-nodejs-skeleton"
                }
             }
          }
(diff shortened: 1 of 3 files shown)


docs/nodejs.md
```diff
@@ -0,0 +1,29 @@
...
+Node.js can be enabled in Local Server by adding `extra.altis.modules.local-server.nodejs` in the project's `composer.json`
+
+```json
+{
+	"extra": {
+		"altis": {
+			"modules": {
+				"local-server": {
+					"nodejs": {
+            "path": "../altis-nodejs-skeleton",
```
joehoyle, line 16: tabs / spaces i think mixed here.

docs/nodejs.md
```diff
@@ -0,0 +1,29 @@
...
+            "version": "21.1"
```
joehoyle, line 17: Can we instead load the version from `${path}/package.json`? That's how we ask clients to specify their node version in Altis Cloud.

inc/composer/class-docker-compose-generator.php
```diff
@@ -237,6 +237,53 @@ protected function get_service_php() : array {
...
+				'volumes' => [
+					"../{$config['nodejs']['path']}/:/usr/src/app"
+				],
+				'working_dir' => '/usr/src/app',
+				'command' => 'npm run dev',
+				'networks' => [
+					'proxy',
+					'default',
+				],
+				'healthcheck' => [
```
joehoyle, line 266: I'm not sure about an HTTP healthcheck. if my node add has a fatal, I don't really want my container to start restarting? Or maybe I do? I don't know, can you explain the usecase for the healthcheck

inc/composer/class-docker-compose-generator.php
```diff
@@ -237,6 +237,53 @@ protected function get_service_php() : array {
...
+	 * @return array
+	 */
+	protected function get_service_nodejs() : array {
+		$config = $this->get_config();
+
+		$version = (string) $this->get_config()['nodejs']['version'] ?? '20';
+
+		return [
+			'nodejs' => [
+				'image' => "node:{$version}-bookworm",
```
joehoyle, line 252: Could we also set env vars for atleast `ALTIS_ENVIRONMENT_NAME` and `ALTIS_ENVIRONMENT_TYPE`, so people can do checks for `local`?

joehoyle:
Really great @jerico!

inc/composer/class-docker-compose-generator.php
jerico: I did not consider those. I just copied how it was implemented on other services we run. Your point makes sense though, since the container will continually watch for file changes and recompile. 

I'll remove it for now.

docs/nodejs.md
jerico: Implemented

inc/composer/class-docker-compose-generator.php
jerico: Implemented

docs/nodejs.md
```diff
@@ -0,0 +1,34 @@
...
+}
+```
+
+`path` refers to the relative path of the project's front-end code.
+
+## Setting Node.js Version
+Similar to configuring the Altis infrastructure, the Local Server determines the Node.js version to use based on the `engines.node` value found in the `package.json` at the specified `path`.
+
+## Running Development Server
+Once configured, the Local Server executes `npm run dev` inside the Node.js container at the specified path. This command watches for changes and recompiles necessary files.
```
mikelittle, line 31: I tried a new `altis-skeleton` with `altis-nodejs-skeleton` in a `nodejs` subdirectory. I had thought the nodejs code must live in the same repos as the Altis code.
Anyway, I saw no evidence of it having run `npm run dev`. 
I tried looking at the nodejs container logs (I had to use vs-code) and it doesn't seem to show `npm run dev` instead it looks like it may have run `next dev`
```sh
added 23 packages, and audited 24 packages in 38s

3 packages are looking for funding
  run `npm fund` for details

1 high severity vulnerability

To address all issues, run:
  npm audit fix

Run `npm audit` for details.

> altis-skeleton@0.0.1 dev
> next dev

   ▲ Next.js 14.0.2
   - Local:        http://localhost:3000

 ✓ Ready in 1886ms
 ○ Compiling / ...
Browserslist: caniuse-lite is outdated. Please run:
  npx browserslist@latest --update-db
  Why you should do it regularly: https://github.com/browserslist/browserslist#browsers-data-updating
 ✓ Compiled / in 2.1s (201 modules)
```
after that, there are only get requests shown when I refresh my browser. Changing index.jsx does nothing.

mikelittle (changes requested):

docs/nodejs.md
mikelittle: Can we get an entry in `composer server logs` for `nodejs`, please?

jerico: > I tried a new altis-skeleton with altis-nodejs-skeleton in a nodejs subdirectory. I had thought the nodejs code must live in the same repos as the Altis code.

Is this decision documented somewhere? I based this off how thehourglass manage their repo. They have front-end and backend on a separate repo.

Our skeleton uses `next.js`, internally it's running the `npm run dev` https://github.com/humanmade/altis-local-server/pull/710/files#diff-6e1aa008277f9688efac7746b347e1e418cfc7b7e36f5bec209e48022d735022R263

I'll check the auto-recompile and adding log entry.

jerico: @mikelittle  
- [x] Auto-recompile and reload working correctly for me.
- [x] Added `nodejs` entry in `composer server logs`

We also support having the nodejs code in the same repository by pointing to it using `nodejs.path`

https://github.com/user-attachments/assets/3851d878-f024-49be-854c-fbced92d7863

mikelittle: > Is this decision documented somewhere? I based this off how thehourglass manage their repo. They have front-end and backend on a separate repo.

I got the node app location information from [this notion page](https://www.notion.so/altis-dxp/Node-js-Support-8b2e3775c68f43f582c23c337b3ee60c?pvs=4#9a936a6e60214b65a9941d57dbd29d32) but it must have moved on. No worries.

The live reload still doesn't work for me. But I'll put that down to my setup.

mikelittle (commented):
Looks good to me. Well done with this.

mikelittle (approved):

docs/nodejs.md
rmccue: The docs are the better place to consult; they're separate repos as we changed that decision based on feedback. It does also make me wonder if we might want the server _separate_ as a result 🤔

jerico: @rmccue Separate server meaning separate from Local Server?

Given our current setup, it makes sense for it to be part of the Local Server for now. When deployed, Node.js runs alongside WordPress. We currently assume that there's always a WordPress instance paired with the front-end in Altis.

PR #722, opened 2024-08-20
Add ability to modify docker-compose configuration

rmccue:
Adds the ability for any Composer package (including the root/project itself) to modify the docker-compose configuration, making it possible for modules to add containers or alter the existing configuration.

This can be used for better modularity and conditional behaviour, such as for optional features like Elasticsearch or Node.js. It also provides the ability for users to customise Docker behaviour fairly directly in their project.

My plan is to transition the ES + Kibana containers to the Enhanced Search package, and then eventually someday remove it from the default bundle, allowing a clearer opt-in for customers who have actually bought it. PR for that incoming soon.

Addresses https://github.com/humanmade/product-dev/issues/1744

### DIFF
--- a/docs/extra-containers.md
+++ b/docs/extra-containers.md
@@ -1,0 +1,64 @@
+---
+order: 100
+---
+# Extra Containers
+
+Local Server provides a series of Docker containers for the base Altis behaviour. Modules and other packages may add additional containers (typically called "sidecar" containers) or alter the container behaviour through the extension system.
+
+
+## How Local Server works
+
+Local Server internally uses Docker Compose to create and manage containers for the Altis environment. When a user runs the `composer server start` command, Local Server dynamically provisions a `docker-compose.yml` file based on the user's preferences. (This file is only regenerated when starting Local Server to avoid conflicts or surprising behaviour for users.)
+
+Other commands such as `composer server logs` are passthrough-style commands, which primarily wrap `docker-compose` commands, and which work with Docker Compose server names.
+
+Domain routing is internally handled using [Traefik Proxy](https://doc.traefik.io/traefik/) and each service can register user-accessible routes using the [label configuration system](https://doc.traefik.io/traefik/providers/docker/).
+
+Extensions register a class which has the ability to filter the `docker-compose.yml` data before it is written out to a file, allowing the extensions to register additional services and containers, or modify the configuration in other ways.
+
+
+## Writing an extension
+
+Any Composer package (including the root package, i.e. the project) can specify an extension class, which is dynamically loaded by Local Server and which receives the `docker-compose.yml` data to filter.
+
+Packages should implement the [`Altis\Local_Server\Composer\Compose_Extension` interface](https://github.com/humanmade/altis-local-server/blob/master/inc/composer/class-compose-extension.php) with the relevant methods.
+
+**Note:** This class will be loaded in the Composer context, not in WordPress, and neither hooks nor the full codebase will be loaded. Notably, functions like `Altis\get_config()` will not return default values as a result.
+
+For example, to add a sidecar service called `foo`:
+
+```php
+namespace MyModule;
+
+use Altis\Local_Server\Composer\{Compose_Extension, Docker_Compose_Generator};
+
+class Local_Server_Extension implements Compose_Extension {
+	protected Docker_Compose_Generator $generator;
+
+	public function set_config( Docker_Compose_Generator $generator, array $args ) : void {
+		$this->generator = $generator;
+	}
+
+	public function filter_compose( array $config ) : array {
+		$config['services']'foo'] = [
+			'container_name' => "{$this->generator->project_name}-foo",
+			'image' => 'hello-world',
+		];
+
+		return $config;
+	}
+```
+
+This class should then be specified in the package's `composer.json` as `extra.altis.local-server.compose-extension`:
+
+```json
+{
+	"extra": {
+		"altis": {
+			"local-server": {
+				"compose-extension": "Altis\\Enhanced_Search\\Local_Server_Extension"
+			}
+		}
+	}
+}
+```
--- a/docs/extra-containers.md
+++ b/docs/extra-containers.md
@@ -50,7 +50,7 @@ class Local_Server_Extension implements Compose_Extension {
 	}
 
 	public function filter_compose( array $config ) : array {
-		$config['services']'foo'] = [
+		$config['services']['foo'] = [
 			'container_name' => "{$this->generator->project_name}-foo",
 			'image' => 'hello-world',
 		];
--- a/docs/extra-containers.md
+++ b/docs/extra-containers.md
@@ -3,27 +3,37 @@ order: 100
 ---
 # Extra Containers
 
-Local Server provides a series of Docker containers for the base Altis behaviour. Modules and other packages may add additional containers (typically called "sidecar" containers) or alter the container behaviour through the extension system.
+Local Server provides a series of Docker containers for the base Altis behaviour. Modules and other packages may add additional
+containers (typically called "sidecar" containers) or alter the container behaviour through the extension system.
 
 
 ## How Local Server works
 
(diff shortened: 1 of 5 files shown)


docs/extra-containers.md
```diff
@@ -0,0 +1,64 @@
...
+
+class Local_Server_Extension implements Compose_Extension {
+	protected Docker_Compose_Generator $generator;
+
+	public function set_config( Docker_Compose_Generator $generator, array $args ) : void {
+		$this->generator = $generator;
+	}
+
+	public function filter_compose( array $config ) : array {
+		$config['services']'foo'] = [
```
kovshenin, line 43:
```suggestion
		$config['services']['foo'] = [
```

docs/extra-containers.md
```diff
@@ -0,0 +1,64 @@
...
+order: 100
+---
+# Extra Containers
+
+Local Server provides a series of Docker containers for the base Altis behaviour. Modules and other packages may add additional containers (typically called "sidecar" containers) or alter the container behaviour through the extension system.
+
+
+## How Local Server works
+
+Local Server internally uses Docker Compose to create and manage containers for the Altis environment. When a user runs the `composer server start` command, Local Server dynamically provisions a `docker-compose.yml` file based on the user's preferences. (This file is only regenerated when starting Local Server to avoid conflicts or surprising behaviour for users.)
```
mikelittle, line 11:
```suggestion
Local Server internally uses Docker Compose to create and manages containers for the Altis environment. When a user runs the `composer server start` command, Local Server dynamically provisions a `docker-compose.yml` file based on the user's preferences. (This file is only regenerated when starting Local Server to avoid conflicts or surprising behaviour for users.)
```

mikelittle: Could you also wrap lines at 132 chars please? As per our MD [linting standards](https://github.com/humanmade/altis-dev-tools-command/blob/db0985a71418a4e11b2b287b30827db4c6d42d0f/templates/docslint/markdownlint.yaml#L61).

inc/composer/class-command.php
```diff
@@ -116,6 +120,40 @@ private function get_base_command_prefix() : string {
...
+		$packages = [];
+		foreach ( $repo->getPackages() as $package ) {
+			$packages[ $package->getName() ] = $package;
+		}
+
+		// Check which are publishing containers.
+		$config = [];
+		foreach ( $packages as $name => $package ) {
+			$extra = $package->getExtra();
+			if ( ! isset( $extra['altis'] ) || ! isset( $extra['altis']['local-server'] ) ) {
```
mikelittle, line 140: This would be clearer as a positive test:
```suggestion
			if ( isset( $extra['altis'] ) && isset( $extra['altis']['local-server'] ) ) {
				$config[ $name ] = $extra['altis']['local-server'];
			}
```

mikelittle (changes requested):
A couple of suggestions.
Also, the lint issues will need fixes (mostly missing docs).

mikelittle:
I like the idea of this. 
It might be an opportunity to stop using the implicit dependency and networking configuration of [`links`](https://docs.docker.com/reference/compose-file/services/#links). Instead, we should use explicit [`depends_on`](https://docs.docker.com/reference/compose-file/services/#depends_on) and [`networks`](https://docs.docker.com/reference/compose-file/services/#networks) attributes.

inc/composer/class-command.php
rmccue: I'd agree if it were a straight `if`/`else`, but this is more of a return-early pattern. Not strongly opinionated so if you feel it's much more readable that way I'll update

kovshenin (approved):

inc/composer/class-command.php
mikelittle: Maybe `empty` would read better. `if (empty( $extra['altis']) || (empty($extra['altis']['local-server']))`

CodeProKid:
@rmccue Just noting that we'd love to have the ability to do modify the docker compose from our own project! Our immediate use case for this would be to prefix all of the image names in the generated docker-compose file with our internal docker hub proxy URL so they get pulled from there instead of directly from docker hub which will prevent us from running into [docker hub API limits](https://docs.docker.com/docker-hub/download-rate-limit/) in CI.

It looks like we'd be able to achieve this quite easily with the way this feature has been designed 😃

mikelittle:
I did test this locally with the ES example. It highlighted a weird error that I think it is a dependency issue, though it doesn't make sense.
With this branch and the [ES PR](https://github.com/humanmade/altis-enhanced-search/pull/525) checked out, I get the following error on startup:
```shell
$ composer server start
Starting...
PHP Fatal error:  Declaration of Symfony\Component\Console\Input\ArrayInput::hasParameterOption(array|string $values, bool $onlyParams = false): bool must be compatible with Symfony\Component\Console\Input\InputInterface::hasParameterOption($values, bool $onlyParams = false) in /Users/mikelittle/sandbox/02-altis/product-dev/vendor/symfony/console/Input/ArrayInput.php on line 50

Fatal error: Declaration of Symfony\Component\Console\Input\ArrayInput::hasParameterOption(array|string $values, bool $onlyParams = false): bool must be compatible with Symfony\Component\Console\Input\InputInterface::hasParameterOption($values, bool $onlyParams = false) in /Users/mikelittle/sandbox/02-altis/product-dev/vendor/symfony/console/Input/ArrayInput.php on line 50
```

Grep-ing for the function gets me three hits—all correct:

```shell
$ grep -r 'hasParameterOption'
... skipping calls
vendor/symfony/console/Input/ArrayInput.php:    public function hasParameterOption(string|array $values, bool $onlyParams = false): bool
vendor/symfony/console/Input/ArgvInput.php:    public function hasParameterOption(string|array $values, bool $onlyParams = false): bool
vendor/symfony/console/Input/InputInterface.php:    public function hasParameterOption(string|array $values, bool $onlyParams = false): bool;
```

So far, I made it work by specifying `"symfony/console": "5.4"` as a requirement in `dev-tools`. I've not committed this yet, But will see if the problem shows up elsewhere.

However, this issue apart, the code did work. I could enable/disable the Elastic Search container from the Enhanced Search module.

joehoyle:
+1 from me on this approach

joehoyle:
@mikelittle I got the same issue as you, it appears that it's actually conflicting with a version of `Console` that is in `composer`. In my case `Input` / `InputInterface` is declared in `phar:///opt/homebrew/Cellar/composer/2.6.5/bin/composer/vendor/symfony/console/Input/Input.php`. it's not clear to my why `ArrayInput` isn't also getting loaded from the composer's deps instead, maybe the class-loader order is unspecified?

joehoyle:
So, Composer (2.6.5) at least is bundling Symfony Console `5.4.28.0`, whereas my altis install is using `6.2.x-dev`, `composer why symfony/console` shows:

```
codeception/codeception 5.1.2      requires  symfony/console (>=4.4.24 <8.0)                                                  
psy/psysh               0.12.x-dev requires  symfony/console (^7.0 || ^6.0 || ^5.0 || ^4.0 || ^3.4)                           
psy/psysh               0.12.x-dev conflicts symfony/console (4.4.37 || 5.3.14 || 5.3.15 || 5.4.3 || 5.4.4 || 6.0.3 || 6.0.4) 
symfony/var-dumper      6.4.x-dev  conflicts symfony/console (<5.4)                                                           
symfony/yaml            6.3.x-dev  conflicts symfony/console (<5.4) 
```

I don't know to what degree the `composer` dependencies are taken into account when resolving versions...

joehoyle:
It appears `composer` doesn't account at all for it's own deps, so it's relatively easy to get a version conflict between a project's deps and ones bundled in `composer.phar`. I'm not sure exactly how that should be handled. I'm assuming this is why we won't specify `symfony/console` as a dep of `local-server`, but it is getting loaded in via dev-tools, so as @mikelittle points out, sticking dev tools to `symfony/console": "5.4"` should resolve this. However, I don't know if other versions of the composer binary will include different major versions of `console`.

mikelittle (approved):
I think this is looking good now and ready to go.

wisyhambolu (commented):
@mikelittle I'm seeing a few linter errors, can these be fixed?

wisyhambolu (approved):
I tested this based on the tests described in https://github.com/humanmade/product-dev/issues/1744#issuecomment-2748723822 and i was able to verify that it works as expected. The Elasticsearch and Kibana containers are no longer created by default.

PR #723, opened 2024-08-20
Move Elasticsearch containers to Search module

rmccue:
Builds off #722 and requires that.

Addresses: https://github.com/humanmade/product-dev/issues/1744


docs/elasticsearch.md
```diff
@@ -1,20 +1,44 @@
-# ElasticSearch
+# Elasticsearch
+
+Elasticsearch is available on some Altis plans, and is fully integrated into Altis with the [Enhanced Search module](docs://search/), enabling enhanced search and relevancy as well as powering the analytics data query layer.
```
mikelittle, line 3: Is the "powering the analytics data query layer." part still true?

mikelittle (commented):
Just a question for now.

mikelittle:
Rebased and force-pushed

docs/elasticsearch.md
rmccue: No, analytics isn't part of Altis proper any more so that needs removing really.

rmccue:
@mikelittle Generally avoid rebasing and use merge commits so that we don't lost the history here.

docs/elasticsearch.md
wisyhambolu: @mikelittle Looks like this still needs resolving?

mikelittle:
The only linter issue left now is a php version issue.

wisyhambolu (approved):
I tested this based on the tests described in https://github.com/humanmade/product-dev/issues/1744#issuecomment-2748723822 and i was able to verify that it works as expected. The Elasticsearch and Kibana containers are no longer created by default.

PR #748, opened 2024-10-24
Revert "Update symfony/yaml requirement from ~5.4.40 to ~7.1.5"

mikelittle:
Reverts humanmade/altis-local-server#740

That merge broke down the line dependencies in elasticpress, etc.


wisyhambolu (approved):
LGTM

PR #763, opened 2024-12-05
Clarify how PHP versions are specified.

mikelittle:
Clarify how the PHP versions need to be specified in `composer.json`

Fixes https://github.com/humanmade/altis-documentation/issues/417

### DIFF
--- a/docs/php-version.md
+++ b/docs/php-version.md
@@ -19,6 +19,9 @@ To change the PHP version use the Local Server module configuration to set the `
 }
 ```
 
+**Note**: Only the major and minor versions (x.y) should be listed. We will use the latest patch version we have tested and made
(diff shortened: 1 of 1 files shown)


wisyhambolu (approved):

docs/php-version.md
```diff
@@ -19,6 +19,9 @@ To change the PHP version use the Local Server module configuration to set the `
 }
 ```
 
+**Note**: Only the major and minor versions (x.y) should be listed. We will use the latest patch version we have tested and made
+available both locally and in the cloud.
```
rmccue, line 23:
```suggestion
**Note**: Only the major and minor versions (x.y) should be specified. Altis automatically manages and updates the patch version
as part of regular maintenance.
```

Issue #745, opened 2024-10-16
Add option to override default site path


rmccue:
For sites where the default site isn't the root of the local URL, it'd be useful to have a way to override the default that we provide to `wp-cli`. e.g. if your default site is at `example.altis.dev/en/`, you currently need to pass the URL every time.

I'd suggest we add a `default_site_url` option (e.g. `"default_site_url": "https://example.altis.dev/en/"`), or a `default_site_path` option (e.g. `"default_site_path": "/en/"`) to override this.

---

## Acceptance Criteria

- [x] When specifying a default site URL as `"default_site_url": "example.altis.dev/en"` in `composer.json`, `composer server cli` commands work without passing `--url`.
- [x] When not explicitly specifying a default site URL in `composer.json`, `composer server cli` commands use a default URL based on the domain setting
- [x] Given the above, CI builds also work
- [x] Passing a `--url` flag for a different sub site still works.
- [x] Tachyon URLs are still generated as `example.altis.dev/tachyon` and continue to function.
- [x] Perform manual testing on this with subsites at `/` and `/en/` to check the default is working. (No automated testing required, as we don't have a framework in place.)
- [x] Record the process of testing into the ticket so that it can be replicated later.
- [x] Document in Local Server's docs about config options.

---

### Ready for Work Checklist

Is this ticket ready to be worked on? See
[the Play Book Definition of Ready.](https://playbook.hmn.md/play/product/definition-of-ready/)

- [x] Is the title clear?
- [x] Is the description clear and detailed enough?
- [x] Are acceptance criteria listed?
- [x] Have any dependencies been identified? (Optional)
- [x] Have any documentation/playbook changes been identified? (Optional)
- [x] Is an estimate or time box assigned?
- [x] Is a priority label assigned?
- [x] Is this ticket added to a milestone?
- [ ] Is this ticket added to an epic? (Optional)

### Completion Checklist

Is this ticket done? See
[the Play Book Definition of Done.](https://playbook.hmn.md/play/product/definition-of-done-2/)

- [ ] Has the acceptance criteria been met?
- [ ] Is the documentation updated (including README)?
- [ ] Do any code/documentation changes meet project standards?
- [ ] Are automatic tests in place to verify the fix or new functionality?
  - [ ] Or are manual tests documented (at least on this ticket)?
- [ ] Are any Playbook/Handbook pages updated?
- [ ] Has a new module release (patch/minor) been created/scheduled?
[...]

PR #773, opened 2025-01-08
Add `default-site-url` override option for WP CLI commands

wisyhambolu:
- Delivers https://github.com/humanmade/altis-local-server/issues/745
- While working on this I realised that when a `wp` command is run via `composer server exec -- wp` the site url set in `--url` does not get used by the wp command, so I added code to enable that in https://github.com/humanmade/altis-local-server/pull/773/commits/6626d547f22ad4cf91073fe65a636535400a0592 to ensure that the `default-site-url` also applies to WP-CLI commands run via `composer server exec -- wp`.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -43,3 +43,26 @@ composer server cli -- db import database.sql
 
 **Note:** For privacy reasons any database backups that are version controlled should have any personally identifiable information
 removed and extra care should be taken to avoid committing database backup files containing personal data.
+
+### Default Site URL
+
+Altis supports specifying a default site URL that can be different from the root URL of the local environment. This is particularly useful in scenarios where the default site resides at a subpath rather than the root of the local URL. By configuring this option, users can ensure WP-CLI commands operate on the correct site by default.
+
+You can define the default site URL in your project’s `composer.json` file under the `extra.altis.modules.local-server.default-site-url` property. Here’s an example:
+
+
+```json
+{
+    "extra": {
+        "altis": {
+            "modules": {
+                "local-server": {
+                    "default-site-url": "my-site.altis.dev/en/"
+                }
+            }
+        }
+    }
+}
+```
+
+Note: By setting `default-site-url`, the default site URL will only be overridden for WP-CLI commands. It does not affect other components or tools in your environment.
--- a/docs/default-site-url.md
+++ b/docs/default-site-url.md
@@ -1,0 +1,22 @@
+# Default Site URL
+
+Altis supports specifying a default site URL that can be different from the root URL of the local environment. This is particularly useful in scenarios where the default site resides at a subpath rather than the root of the local URL. By configuring this option, users can ensure WP-CLI commands operate on the correct site by default.
+
+You can define the default site URL in your project’s `composer.json` file under the `extra.altis.modules.local-server.default-site-url` property. Here’s an example:
+
+
+```json
+{
+    "extra": {
+        "altis": {
+            "modules": {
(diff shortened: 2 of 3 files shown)


mikelittle:
@wisyhambolu I found some issues when testing this

Having set `alts.modules.local-server.default-site-url` to "altis-dev.altis.dev/en" and then run `composer server start` a couple of things were noticeable.
1. At the end of the composer command it just mentions "To access your site visit: https://altis-dev.altis.dev/" without the subdirectory.
2. Although accessing "https://altis-dev.altis.dev/en/" works to get the root site, so does "https://altis-dev.altis.dev/"
3. It looks like it has set the `home` and `siteurl` options in `wp_options` correctly, so `wp site list` gives the correct output. But looking in the `wp_blogs` table it shows a path of '/' for the root site. 
CleanShot 2025-01-09 at 17 09 00@2x
I'm not 100% sure that is correct. Though, it may be needed to make the admin site of things work. I haven't investigated yet.
4. Further to 2 above, all urls to posts and page on the `/en` site also work without the `/en`. I think that's defineitely a problem. I will invesirtgate whether pure WordPress supports this correctly.

wisyhambolu:
@mikelittle Thanks for the review. 

> At the end of the composer command it just mentions "To access your site visit: https://altis-dev.altis.dev/" without the subdirectory.

I think this is the expected behaviour. From my understanding of the issue and the acceptance criteria, the scope of usage of the value set in `default-site-url` is WP-CLI commands only and not in the UI or DB so I didn't make any changes to the `site_url` globally.

> For sites where the default site isn't the root of the local URL, it'd be useful to have a way to override the default that we provide to wp-cli. e.g. if your default site is at example.altis.dev/en/, you currently need to pass the URL every time.

The description of the issue mentions `overriding the default that we provide to wp-cli` which I understand to be scoping this to just the URL that WP-CLI uses, saving the user the burden of always typing --url when using the `composer server cli -- ` command. The acceptance criteria also support the same thought which is why the changes were only scoped to WP-CLI commands intentionally but from your comments it seems expectations may be differing 😅 or maybe I misunderstood the description/acceptance criteria.

> Although accessing "https://altis-dev.altis.dev/en/" works to get the root site, so does "https://altis-dev.altis.dev/"

This behaviour is different for me, access https://platform-test.altis.dev/ gives me the root site and https://platform-test.altis.dev/en/ gives me a different site.

image

image

I'm happy to expand the scope if needed, I think we'll also need to update the acceptance criteria to be clear about that.

jerico (approved):
This looks good to me given the scope only applies to local server's WP CLI, it checks all the acceptance criteria.

I'm curious if `default_site_path` is intentionally not included?

humanmade-cloud (approved):
I agree that the issues I found are out of scope for this ticket. I think the original client site that spurred this ticket has other measures in place that resolve the issues I found. 
Happy to go with the functionality as is. 
I do agree with Joe that the documentation probably works better on this page https://docs.altis-dxp.com/local-server/cli/

wisyhambolu:
Thanks for the suggestions @joehoyle and @mikelittle, the doc has been moved to `cli.md`

wisyhambolu:
> This looks good to me given the scope only applies to local server's WP CLI, it checks all the acceptance criteria.
> 
> I'm curious if `default_site_path` is intentionally not included?

@jerico Yes, it was intentionally not included because the issue initially specified implementing either `default-site-url` or `default_site_path` but the decision was later to go with `default-site-url` which is why `default_site_path` wasn't included.

PR #778, opened 2025-01-31
Name and autoremove the import-uploads command's minio container

kadamwhite:
This prevents Docker from becoming littered with repeated "lovable_badger", "terrible_matsumoto", "mysterious_platypus" leftover default-name for the minio containers:
image

Changes:

- Clean up the container after Exit
- Name it based on the project domain it so that it's more easily identifiable in Docker tooling while running


joehoyle (approved):
Makes sense, thanks @kadamwhite !

Issue #712, opened 2024-07-18
Document the New Log Aliases for 'db'


mikelittle:
[This Issue](https://github.com/humanmade/altis-local-server/issues/272) and [PR](https://github.com/humanmade/altis-local-server/pull/691) added Aliases for the `composer server logs db` command.

The documentation on [this page ](https://docs.altis-dxp.com/local-server/viewing-logs/) should be updated to match.

---

## Acceptance Criteria

- [ ] The new aliases are documented

---

### Ready for Work Checklist

Is this ticket ready to be worked on? See
[the Play Book Definition of Ready](https://playbook.hmn.md/play/product/definition-of-ready/)

- [x] Is the title clear?
- [x] Is the description clear and detailed enough?
- [x] Are acceptance criteria listed?
- [x] Have any dependencies been identified? (Optional)
- [x] Have any documentation/playbook changes been identified? (Optional)
- [x] Is an estimate or time box assigned?
- [x] Is a priority label assigned?
- [x] Is this ticket added to a milestone?
- [ ] Is this ticket added to an epic? (Optional)

### Completion Checklist

Is this ticket done? See
[the Play Book Definition of Done](https://playbook.hmn.md/play/product/definition-of-done-2/)

- [ ] Has the acceptance criteria been met?
- [ ] Is the documentation updated (including README)?
- [ ] Do any code/documentation changes meet project standards?
- [ ] Are automatic tests in place to verify the fix or new functionality?
  - [ ] Or are manual tests documented (at least on this ticket)?
- [ ] Are any Playbook/Handbook pages updated?
- [ ] Has a new module release (patch/minor) been created/scheduled?
- [ ] Have the appropriate `backport` labels been added to the PR?

PR #792, opened 2025-02-13
Document the new logs aliases for db.

mikelittle:
Adds documentation for the aliases for the `composer server logs db` command.

Addresses https://github.com/humanmade/altis-local-server/issues/712


ferschubert-hm (approved):
Looks good.

Issue #818, opened 2025-05-08
Implement default + configurable memory limits for docker containers


svandragt:
Steps to reproduce:
1. Check out an Altis project on a linux device
2. `composer server start && docker stats`
3. Observer apart from the ES container, all other containers have a memory limit of the local machine's installed memory size

This happens because Docker Desktop is not setting an upper bound, it's not used. I suppose it would be the same on other platforms when DD is not used, not just linux, but I haven't tested this.

This causes the local machine to hang when running out of memory, in my case due to to `minio/mc` taking up 19+GB watching the local filesystem for changes to sync to s3 until the desktop environment became unresponsive.

Sensible limits for all containers.

I've enclosed a patch that sets 1G limits for containers if none are specified:
<details>
<summary>Click to toggle contents of `diff`</summary>

```diff

Index: vendor/altis/local-server/inc/composer/class-docker-compose-generator.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/vendor/altis/local-server/inc/composer/class-docker-compose-generator.php b/vendor/altis/local-server/inc/composer/class-docker-compose-generator.php
--- a/vendor/altis/local-server/inc/composer/class-docker-compose-generator.php
+++ b/vendor/altis/local-server/inc/composer/class-docker-compose-generator.php
@@ -225,20 +217,31 @@
 		return $services;
 	}

+	private function apply_service_defaults( $args ) : array {
+		foreach ( $args as $service => $service_args ) {
+			if ( isset( $service_args['mem_limit'] ) ) {
+				continue;
+			}
+			$args[ $service ]['mem_limit'] = '1g';
+		}
+
+		return $args;
+	}
+
 	/**
 	 * Get the PHP container service.
 	 *
 	 * @return array
 	 */
 	protected function get_service_php() : array {
-		return [
+		return $this->apply_service_defaults( [
 			'php' => array_merge(
 				[
 					'container_name' => "{$this->project_name}-php",
 				],
 				$this->get_php_reusable()
 			),
-		];
+		] );
 	}

 	/**
@@ -253,7 +256,7 @@
 		$package_json = json_decode( file_get_contents( "{$config['nodejs']['path']}/package.json" ), true );
 		$version = $package_json['engines']['node'] ?? '20';

-		return [
+		return $this->apply_service_defaults( [
 			'nodejs' => [
[...]

PR #819, opened 2025-05-08
 Limit container memory

mikelittle:
Work in progress

 Implement a default memory limit for containers if one is not specified.
 For now it applies the same limit across all services.
 But as enhanced-search now specifies its own docker-compose config, that is not affected by this. The es contaainer has a memry limit but the kibana container does not.
 Similarly, the proxy container is independent and does not specify a limit.

 Implemements https://github.com/humanmade/altis-local-server/issues/818

### DIFF
--- a/inc/composer/class-docker-compose-generator.php
+++ b/inc/composer/class-docker-compose-generator.php
@@ -226,19 +226,39 @@ class Docker_Compose_Generator {
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


svandragt (approved):
Linttert issues not withstanding, looks good.

svandragt:
Hi is it possible to backport this to supported altis versions?

PR #850, opened 2025-09-15
Automatically roll SSL certs

### LINKED ISSUE #849: Automatically regenerate SSL certificates when expired
### Context

When starting Local Server after running a project for a long time, the certificate may be expired. This leads to browser errors like `SEC_ERROR_EXPIRED_CERTIFICATE`

We know enough about the certificates to be able to determine this and automatically fix it for the users.

### Current Behaviour

Starting the server once the certificate has expired leads to certificate errors in the browser. Users need to run `composer server ssl generate` manually.

There is also no explanation of the problem or how to solve it immediately available to users.

### Improved Behaviour

We should automatically regenerate expired certificates on `composer server start`. Additionally, the `composer server ssl` command should indicate how long is left on the expiration.

### Impact

-

### Possible Solutions

_No response_

### Acceptance Criteria

_No response_

rmccue:
Automatically regenerates certificates on `composer server start` if they're expiring soon:
Screenshot 2025-09-15 at 15 22 30

Also changes `composer server ssl` to output remaining validity days:
```
$ composer server ssl
Certificate exists and is valid for 822 more days
```

And error if it's expired.

Fixes #849.

### DIFF
--- a/inc/composer/class-command.php
+++ b/inc/composer/class-command.php
@@ -14,6 +14,7 @@ namespace Altis\Local_Server\Composer;
 
 use Composer\Command\BaseCommand;
 use Composer\Composer;
+use Exception;
 use Symfony\Component\Console\Input\ArrayInput;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
@@ -1032,7 +1033,7 @@ EOT;
 	/**
 	 * Get the SSL certificate expiration date.
 	 *
-	 * @throws Exception if openssl is missing, the certificate is invalid, or it cannot be read.
+	 * @throws Exception If openssl is missing, the certificate is invalid, or it cannot be read.
 	 * @param string $path Path to the certificate file.
 	 * @return int Days until expiry. If negative, the certificate has already expired.
 	 */


rmccue:
Remaining lint error is due to a PHP incompatibility, and is not part of this PR.

inc/composer/class-command.php
```diff
@@ -971,15 +983,22 @@ protected function ssl( InputInterface $input, OutputInterface $output ) {
...
 				if ( ! $cert_exists ) {
 					$output->writeln( "<error>Certificate file does not exist. Use 'composer server ssl generate' to generate one. </error>" );
 					return 1;
-				} else {
-					$output->writeln( '<info>Certificate file exists.</info>' );
 				}
 
-				break;
+				// Check certificate expiration.
+				$validity_remaining = $this->check_ssl_expiry( $cert_file );
```
joehoyle, line 994: Hmm are we catching this error? Do we want to block being able to start the server if the `openssl` call fails? IMO best to just show a warning "Could not get certificate expiry details" but still start the server?

inc/composer/class-command.php
```diff
@@ -971,15 +983,22 @@ protected function ssl( InputInterface $input, OutputInterface $output ) {
 				break;
 
 			case '':
-				$cert_exists = file_exists( 'vendor/ssl-cert.pem' ) && file_exists( 'vendor/ssl-key.pem' );
+				$cert_file = 'vendor/ssl-cert.pem';
+				$cert_exists = file_exists( $cert_file ) && file_exists( $cert_file );
```
mikelittle, line 987: Should this second test be for the key file?

mikelittle (changes requested):

PR #899, opened 2026-03-03
Support private uploads: add TACHYON_SERVER_VERSION

mikelittle:
## Summary

- Define `TACHYON_SERVER_VERSION` constant (`3.0.0`) when Tachyon is enabled, enabling presigned URL passthrough for private media

## Context

This change supports the default-private uploads feature in the media module. The local Tachyon image (v3.0.7+) supports presigned URL passthrough, but the constant was not being set in local-server.

Ref: https://github.com/humanmade/altis-media/issues/162

### DIFF
--- a/inc/composer/class-docker-compose-generator.php
+++ b/inc/composer/class-docker-compose-generator.php
@@ -586,9 +586,7 @@ class Docker_Compose_Generator {
 					'/bin/sh',
 					'-c',
 					sprintf(
-						"aws s3api create-bucket --bucket %s --endpoint-url=http://s3:7070 || true; aws s3api put-bucket-policy --bucket %s --policy '{\"Version\":\"2012-10-17\",\"Statement\":[{\"Sid\":\"PublicReadUploads\",\"Effect\":\"Allow\",\"Principal\":\"*\",\"Action\":[\"s3:GetObject\"],\"Resource\":[\"arn:aws:s3:::%s/uploads/*\"]}]}' --endpoint-url=http://s3:7070 || true",
-						$this->bucket_name,
-						$this->bucket_name,
+						"aws s3api create-bucket --bucket %s --endpoint-url=http://s3:7070 || true",
 						$this->bucket_name
 					),
 				],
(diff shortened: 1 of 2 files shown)


inc/composer/class-docker-compose-generator.php
```diff
@@ -586,9 +586,7 @@ protected function get_service_s3() : array {
 					'/bin/sh',
 					'-c',
 					sprintf(
-						"aws s3api create-bucket --bucket %s --endpoint-url=http://s3:7070 || true; aws s3api put-bucket-policy --bucket %s --policy '{\"Version\":\"2012-10-17\",\"Statement\":[{\"Sid\":\"PublicReadUploads\",\"Effect\":\"Allow\",\"Principal\":\"*\",\"Action\":[\"s3:GetObject\"],\"Resource\":[\"arn:aws:s3:::%s/uploads/*\"]}]}' --endpoint-url=http://s3:7070 || true",
-						$this->bucket_name,
-						$this->bucket_name,
+						"aws s3api create-bucket --bucket %s --endpoint-url=http://s3:7070 || true",
```
jerico: IIRC I tried it this way but had access issue with fresh/existing installation. I might have remembered it incorrectly. All working with private uploads off after this change?

mikelittle: I  thought so, but this was long enough ago, I need to recheck.

mikelittle: Yes. You are right. With the private uploads feature turned off, this breaks uploads. I'll revert that part. So the only thing left is the setting of the correct tachyon version.

jerico: I think it's worth exploring how to make it work both ways (if possible!). Developers could test private media locally.

mikelittle: The point is they can't. Because our S3 emulation ignores per object ACLs you cannot actually make it work locally.

jerico (approved):

PR #900, opened 2026-03-04
Fix cases where `$PWD` is empty in `Process()`

roborourke:
## Describe the PR 

I haven't managed to ascertain exactly why but the value of `$PWD` in the proxy.yml file was not resolving correctly in an Altis v25 project. I couldn't find anything specific about a change in behaviour in this regard, but these changes seem to make it more consistent and robust.

The error was that the volume mounts were failing, creating empty directories for the `.toml` file, and certificate files in the project root, rather than based from `vendor`, despite it being set as the 2nd param to `new Process()`.

---

### Completion Checklist
Does this PR meet our definition of done? See [the Play Book Definition of Done](https://playbook.hmn.md/play/product/definition-of-done-2/)

- [ ] Has the acceptance criteria been met?
- [ ] Is the documentation updated (including README)?
- [ ] Do any code/documentation changes meet project standards?
    - [ ] OR, are manual tests documented (at least on the ticket/PR)?
- [ ] Are any Playbook/Handbook pages updated?
- [ ] Has a new module release (patch/minor) been created/scheduled?
- [ ] Have the appropriate `backport` labels been added to the PR?
- [ ] Is there a roll-out (and roll-back) plan, if required?

### DIFF
--- a/docker/proxy.yml
+++ b/docker/proxy.yml
@@ -6,10 +6,10 @@ services:
     container_name: altis-proxy
     mem_limit: 128m
     volumes:
-      - "$PWD/altis/local-server/docker/conf/traefik.toml:/etc/traefik/traefik.toml"
-      - "$PWD/altis/local-server/docker/conf/dynamic.toml:/etc/traefik/dynamic.toml"
-      - "$PWD/ssl-cert.pem:/etc/traefik/ssl-cert.pem"
-      - "$PWD/ssl-key.pem:/etc/traefik/ssl-key.pem"
+      - "$PWD/vendor/altis/local-server/docker/conf/traefik.toml:/etc/traefik/traefik.toml"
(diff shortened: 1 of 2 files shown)


mikelittle:
The root cause is that since Composer 2.3, the `process()` method (line 1231) converts commands from strings to arrays. When PHP's `proc_open` receives an array, it executes the command directly without invoking a shell (see PHP docs). Without a shell, the PWD environment variable is not updated to match the `cwd` parameter — it retains the parent process's value. Docker Compose interpolates `$PWD` in proxy.yml from the environment, so it was resolving volume mount paths against the wrong directory. This PR resolves the issue by using fully qualified paths from the project root and removing the `cwd` override, so `$PWD` and the actual working directory are always consistent.

mikelittle (approved):
Approved.
I can't reproduce the issue locally any more, but I know I have hit it in the past and others more recently.

PR #932, opened 2026-05-13
Bump Traefik proxy image to 3.7.1

mikelittle:
## Summary

- Upgrades the Local Server proxy container from `traefik:3.6.12` to `traefik:3.7.1`.
- Picks up the CVE-2026-44774 fix (also in 3.6.17) and the CVE-2026-41181 fix from 3.6.15.
- 3.6 → 3.7 breaking changes are limited to the Kubernetes Gateway API / CRD providers, which Local Server does not use. The Docker provider's documented Docker Engine minimum (≥ v19.03 / API v1.40) is unchanged from 3.6.16. All existing route rules (`Host`, `HostRegexp`, `PathPrefix`, `replacepathregex`, `stripprefix`, `customrequestheaders`) continue to work unchanged.

## Test plan

- [x] `composer server start` boots cleanly on a fresh checkout.
- [x] Primary host (`<project>.altis.dev`) and a subdomain multisite child (`sub.<project>.altis.dev`) both resolve.
- [ ] `/uploads/...`, `/tachyon/...`, `/webgrind`, `/mailhog` paths still route correctly.
- [x] Traefik dashboard reachable on `:8080`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #933, opened 2026-05-13
Use wildcard Host matchers instead of HostRegexp

mikelittle:
## Summary

Replaces `HostRegexp` route rules in `Docker_Compose_Generator` with Traefik 3.7's single-level wildcard `Host` matcher:

- `get_primary_host_rule()` — `HostRegexp(\`^[A-Za-z0-9-]+\.<host>$\`)` → `Host(\`*.<host>\`)`, for the primary domain and every extra domain.
- `get_s3_client_host_rule()` — same swap for the S3 client subdomain rule.
- nodejs router rule — `HostRegexp(\`nodejs-<host>\`)` (no regex metacharacters) → `Host(\`nodejs-<host>\`)`.

Why:

- Simpler, cheaper matcher (no regex compile).
- TLSOptions can be associated with wildcard `Host` matchers, but **not** with `HostRegexp` (per the 3.7 migration notes). This unblocks per-domain TLS options in the future.

Behavior note: the wildcard accepts any DNS-valid label, while the old regex required `[A-Za-z0-9-]+`. For local dev hostnames this is a deliberate relaxation, not a regression.

## Dependencies

Requires Traefik ≥ 3.7. **Merge #932 first** (or merge them together).

## Test plan

- [ ] Primary host (`<project>.altis.dev`) still resolves.
- [ ] Subdomain multisite child (`sub.<project>.altis.dev`) still resolves.
- [ ] `extra_domains` configured in `local-server` config still resolve at both bare and subdomain levels.
- [ ] `/uploads/...` (S3 client routing) still works via Tachyon and direct.
- [ ] nodejs service (`nodejs-<project>.altis.dev`) routes correctly.

### DIFF
--- a/inc/composer/class-docker-compose-generator.php
+++ b/inc/composer/class-docker-compose-generator.php
@@ -301,7 +301,7 @@ class Docker_Compose_Generator {
 				'labels' => [
 					'traefik.enable=true',
 					'traefik.docker.network=proxy',
-					"traefik.http.routers.{$this->project_name}-nodejs.rule=HostRegexp(`nodejs-{$this->hostname}`)",
+					"traefik.http.routers.{$this->project_name}-nodejs.rule=Host(`nodejs-{$this->hostname}`)",
 					"traefik.http.routers.{$this->project_name}-nodejs.priority=1",
 					"traefik.http.routers.{$this->project_name}-nodejs.entrypoints=web,websecure",
 					"traefik.http.routers.{$this->project_name}-nodejs.service={$this->project_name}-nodejs",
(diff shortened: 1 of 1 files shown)


ferschubert-hm (approved):
Testing Notes

  Tested against a fresh Altis 26 skeleton install. Here's what was needed to get the test environment running and results for each
   test plan item.

  Setup

  To test the branch, added a VCS repository and branch alias to composer.json:

  "repositories": [
      { "type": "vcs", "url": "https://github.com/humanmade/altis-local-server" }
  ],
  "require-dev": {
      "altis/local-server": "dev-traefik-wildcard-host-rules as 26.0.7"
  }

  Two fixes were needed on the branch before composer update succeeded:

  1. symfony/yaml constraint — the branch requires ~6.3.12 but projects on 6.4.x will fail to resolve. Suggest widening to ~6.3.12
  || ^6.4:
  "symfony/yaml": "~6.3.12 || ^6.4"
  2. Traefik version — PR #932 must be merged first or together with this one. With Traefik 3.6.12 still running, the wildcard
  Host(*.x.altis.dev) rules silently don't match and return a plain-text 404 from the proxy. After bumping to 3.7.1 (as per #932)
  everything works as expected. The dependency is already noted in the PR description but worth flagging as a hard blocker for
  testing.

  Test Results

  - ✅ Primary host (altis-pr-933.altis.dev) resolves correctly
  - ✅ Subdomain multisite child (test.altis-pr-933.altis.dev) resolves correctly — confirmed via Traefik dashboard showing
  Host(*.altis-pr-933.altis.dev) matching the subdomain
  - ✅ extra_domains — added extra-test.altis.dev via the domains config key; request reaches WordPress (confirmed by WordPress
  302, not a Traefik 404)
  - ✅ /uploads/ — routes correctly to the S3 service (server: VERSITYGW in response headers)
  - ✅ nodejs (nodejs-altis-pr-933.altis.dev) — routes correctly after enabling the nodejs service

  Behaviour note confirmed

  The wildcard Host(*.x.altis.dev) is a deliberate relaxation vs the old HostRegexp([A-Za-z0-9-]+\.x\.altis\.dev) — for local dev
  this is fine and unblocks future TLS options work.

  Approved

PR #971, opened 2026-07-02
Removing MySQL 8.0 as default from local server

filter182:
- Delivera https://github.com/humanmade/product-dev/issues/2185

### DIFF
--- a/inc/composer/class-docker-compose-generator.php
+++ b/inc/composer/class-docker-compose-generator.php
@@ -452,7 +452,6 @@ class Docker_Compose_Generator {
 	protected function get_service_db() : array {
 		$version_map = [
 			'8.4' => 'mysql:8.4.6',
-			'8.0' => 'mysql:8.0.44',
 		];
 
 		$versions = array_keys( $version_map );
@@ -476,7 +475,7 @@ class Docker_Compose_Generator {
 				'image' => $image,
 				// Suppress mysql_native_password deprecation warning
 				// Only affects in-place upgrades from MySQL 5.7 to 8.0.
-				'command' => $version === '8.0' ? '--log-error-suppression-list=MY-013360' : '',
+				//'command' => $version === '8.0' ? '--log-error-suppression-list=MY-013360' : '',
 				'container_name' => "{$this->project_name}-db",
 				'volumes' => [
 					'db-data:/var/lib/mysql',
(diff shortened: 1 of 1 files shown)


inc/composer/class-docker-compose-generator.php
```diff
@@ -476,7 +475,7 @@ protected function get_service_db() : array {
 				'image' => $image,
 				// Suppress mysql_native_password deprecation warning
 				// Only affects in-place upgrades from MySQL 5.7 to 8.0.
-				'command' => $version === '8.0' ? '--log-error-suppression-list=MY-013360' : '',
+				//'command' => $version === '8.0' ? '--log-error-suppression-list=MY-013360' : '',
```
wisyhambolu, line 478: This should not be commented out but rather removed.

inc/composer/class-docker-compose-generator.php
```diff
@@ -476,7 +475,7 @@ protected function get_service_db() : array {
 				'image' => $image,
 				// Suppress mysql_native_password deprecation warning
 				// Only affects in-place upgrades from MySQL 5.7 to 8.0.
```
wisyhambolu, line 477: These need to be removed also as they a re 8.0 specific.

wisyhambolu (changes requested):
A few comments need to be removed. We should avoid commenting out code as it creates noise within the codebase.

Also, `8.0` references need to be removed in https://github.com/humanmade/altis-local-server/blob/master/docs/database.md also which are missed from this PR.

mikelittle:
@filter182 JC, One thing that is always useful is to connect a PR to the ticket it addresses via the Zenhub widget. You'll need the Zenhub browser extension or view the PR in Zenhub.
SCR-20260709-mpak

It makes a direct link to the PR from the Zenhub board.
Before:
SCR-20260709-mowa

After:
SCR-20260709-mpht

inc/composer/class-docker-compose-generator.php
filter182: I'll close this PR as you mentioned in the ticket that this should be a separated PRs, doing this on that PR

filter182:
> @filter182 JC, One thing that is always useful is to connect a PR to the ticket it addresses via the Zenhub widget. You'll need the Zenhub browser extension or view the PR in Zenhub. SCR-20260709-mpak
> 
> It makes a direct link to the PR from the Zenhub board. Before: SCR-20260709-mowa
> 
> After: SCR-20260709-mpht

@mikelittle I think I did that but I was getting the message that I needed to re-auth that extension which it seems that it didn't work, also in other instances I have the issue that it doesn't allow me to liking because it simply doesn't show the ticket to be lined with

filter182:
closing this in favor of https://github.com/humanmade/altis-local-server/pull/985 and https://github.com/humanmade/altis-local-server/pull/986

PR #981, opened 2026-07-03
Disable Docker CLI hints in Local Server commands

altis-assistant:
## Summary
- disable Docker CLI hints for Local Server commands by exporting `DOCKER_CLI_HINTS=false` at command execution time
- pass `DOCKER_CLI_HINTS=false` through explicit Docker Compose process environments
- include the variable in shell command prefixes used by Local Server helpers

## Testing
- `git diff --check`
- Unable to run PHP syntax checks in this container because `php` is not installed

### DIFF
--- a/inc/composer/class-command.php
+++ b/inc/composer/class-command.php
@@ -123,7 +123,7 @@ EOT
 	 */
 	private function get_base_command_prefix() : string {
 		return sprintf(
-			'cd %s; VOLUME=%s COMPOSE_PROJECT_NAME=%s',
+			'cd %s; DOCKER_CLI_HINTS=false VOLUME=%s COMPOSE_PROJECT_NAME=%s',
 			'vendor',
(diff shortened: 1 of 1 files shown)


mikelittle (changes requested):
These changes only take effect for the `process()` calls that also pass `$this->get_env()`. 
Five don't (312, 419, 469, 500, 1122), so they get no environment at all — and 312 is `-f proxy.yml up -d`, the first thing `server start` runs and exactly where the hints would show. Confirmed by shimming `docker` on PATH to dump the env it receives.

Those calls have never passed `VOLUME`/`COMPOSE_PROJECT_NAME` either, which may be deliberate for a shared proxy — so probably just the hints var rather than `get_env()` wholesale.

Minor: the `get_base_command_prefix()` hunk is redundant. Both its call sites go through `shell_exec`/`passthru`, which already inherit the `putenv()`.

PR #985, opened 2026-07-14
Changing MySQL default from 8.0.x to 8.4.x in local server

filter182:
- Delivers https://github.com/humanmade/product-dev/issues/2185 (partially)

### DIFF
--- a/docs/database.md
+++ b/docs/database.md
@@ -2,26 +2,7 @@
 
 ## Available Versions
 
-Altis supports MySQL versions `8.4` and `8.0`.
-The default version is `8.0`.
-
-To change the MySQL version use the Local Server module configuration to set the `mysql` property:
-
-```json
-{
-    "extra": {
-        "altis": {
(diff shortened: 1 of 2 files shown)


wisyhambolu (commented):
https://github.com/humanmade/altis-local-server/blob/master/docs/database.md needs to be updated to state the 8.4 is the only and default version.

filter182:
@wisyhambolu that is done, forgot to include the docs in this PR I though different branch would be better but included in this one.

mikelittle (approved):
I was going to suggest we remove the mapping to 8.0, but as we are doing it in the current version, it's better to leave it for now, so picking up this change won't break existing v27 builds. ~~But we should make sure to create a ticket to do the removal for the next version (i.e. against master)~~ I see there is already a ticket and PR for that.
Note: this change needs to be backported to v27 so it needs the 'backport v27-branch' label.

PR #986, opened 2026-07-14
Removing MySQL 8.0.X as option in Local Server

filter182:
- Delivers https://github.com/humanmade/product-dev/issues/2185 (partially)
- This will be merged till the end of the spring to be available on V27

### DIFF
--- a/inc/composer/class-docker-compose-generator.php
+++ b/inc/composer/class-docker-compose-generator.php
@@ -885,7 +885,7 @@ class Docker_Compose_Generator {
 			'xray' => $modules['cloud']['xray'] ?? true,
 			'ignore-paths' => [],
 			'php' => '8.4',
-			'mysql' => '8.0',
+			'mysql' => '8.4',
 			'nodejs' => $modules['nodejs'] ?? false,
 		];
 
--- a/inc/composer/class-docker-compose-generator.php
+++ b/inc/composer/class-docker-compose-generator.php
@@ -452,7 +452,6 @@ class Docker_Compose_Generator {
 	protected function get_service_db() : array {
 		$version_map = [
 			'8.4' => 'mysql:8.4.6',
-			'8.0' => 'mysql:8.0.44',
 		];
 
 		$versions = array_keys( $version_map );
@@ -474,10 +473,7 @@ class Docker_Compose_Generator {
 		return $this->apply_service_defaults( [
 			'db' => [
 				'image' => $image,
-				// Suppress mysql_native_password deprecation warning
-				// Only affects in-place upgrades from MySQL 5.7 to 8.0.
-				'command' => $version === '8.0' ? '--log-error-suppression-list=MY-013360' : '',
-				'container_name' => "{$this->project_name}-db",
+			        'container_name' => "{$this->project_name}-db",
 				'volumes' => [
 					'db-data:/var/lib/mysql',
 				],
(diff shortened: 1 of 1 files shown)


inc/composer/class-docker-compose-generator.php
```diff
@@ -474,10 +473,7 @@ protected function get_service_db() : array {
 		return $this->apply_service_defaults( [
 			'db' => [
 				'image' => $image,
-				// Suppress mysql_native_password deprecation warning
-				// Only affects in-place upgrades from MySQL 5.7 to 8.0.
-				'command' => $version === '8.0' ? '--log-error-suppression-list=MY-013360' : '',
-				'container_name' => "{$this->project_name}-db",
+			        'container_name' => "{$this->project_name}-db",
```
wisyhambolu: @filter182 the array formatting here is misaligned and needs to be fixed, see the lint errors in the "Files changes" tab.

wisyhambolu (changes requested):
Like I mentioned previously, the `8.0` references also need to be removed in https://github.com/humanmade/altis-local-server/blob/master/docs/database.md.

Also, the lint errors need to be fixed.

inc/composer/class-docker-compose-generator.php
filter182: @wisyhambolu That is fixed now and the reference you mentioned wad done in the following branch in this PR https://github.com/humanmade/altis-local-server/pull/985/commits/a336fdfc73196336ff821ea78a125161f04aacfa I mistakenly took the wrong branch locally to fix it but it is already merge but it should be good

inc/composer/class-docker-compose-generator.php
```diff
@@ -474,10 +473,7 @@ protected function get_service_db() : array {
 		return $this->apply_service_defaults( [
 			'db' => [
 				'image' => $image,
-				// Suppress mysql_native_password deprecation warning
-				// Only affects in-place upgrades from MySQL 5.7 to 8.0.
-				'command' => $version === '8.0' ? '--log-error-suppression-list=MY-013360' : '',
-				'container_name' => "{$this->project_name}-db",
+			    'container_name' => "{$this->project_name}-db",
```
wisyhambolu: As discussed, you need to fix this linter error by using tabs for indentation instead of spaces.

wisyhambolu (changes requested):

wisyhambolu:
@filter182 I've seen your indentation changes which look good.

We should also remove the `8.0` reference in https://github.com/humanmade/altis-local-server/blob/master/docs/database.md

it still shows `Version: 8.0` under the `composer server db info` section, it should have been removed in https://github.com/humanmade/altis-local-server/pull/985 but it looks like that was missed.

filter182:
> @filter182 I've seen your indentation changes which look good.
> 
> We should also remove the `8.0` reference in https://github.com/humanmade/altis-local-server/blob/master/docs/database.md
> 
> it still shows `Version: 8.0` under the `composer server db info` section, it should have been removed in #985 but it looks like that was missed.

@wisyhambolu yup I totally missed that reference, this is being fixed on https://github.com/humanmade/altis-local-server/pull/996

wisyhambolu (approved):

wisyhambolu:
Looks good now, thanks for making the changes.

PR #987, opened 2026-07-21
Update nginx container

mikelittle:
Updates the nginx container to version 3.7.4

Tested manually, with routing checkes and subdomains.


wisyhambolu (approved):
Looks good

